PR #65, opened 2018-04-03
Improve error handling on manual server entry

dborkan:
Improve error handling on manual server entry
* Show an error message if the user clicks Done when they have not entered any manual server configuration, rather than the feedback screen.  I originally wanted to have the Done button desensitized until text was entered, however this screen uses a native HTML `textarea` rather than a Polymer component (we had trouble styling multiline Polymer inputs).  I tried all different event handlers I could think of, however I could not get events to fire with `this` being the Polymer template.  I also tried using 2-way-data-binding on the `value` attribute but that didn't work either.
* Removes the text "Invalid server configuration: could not parse JSON." from the pre-populated feedback.  This now appears in the sentry logs, but not in the sentry message.


fortuna:
Have you tried `textarea.enabled = false`?

PR #81, opened 2018-04-09
Retry droplet creation on DigitalOcean account not-yet-finalized error

dborkan:
Retry droplet creation on DigitalOcean account not-yet-finalized error.  The logic here is that we attempt to create a droplet, and if it fails with the `ACCOUNT_NOT_YET_FINALIZED_ERROR` message, we retry again in 5 seconds, up to 10 times.  Any other error stills fails as before.

Currently this has been tested by temporarily modifying the droplet request to fail, and temporarily modifying the error string we search for, then verifying that the retry logic works as expected.  We should not merge this in until we can test for real with the new error.

### DIFF
--- a/src/server_manager/cloud/digitalocean_api.ts
+++ b/src/server_manager/cloud/digitalocean_api.ts
@@ -148,9 +148,12 @@ class RestApiSession implements DigitalOceanSession {
             })
             .then(fulfill)
             .catch((e) => {
-              const ABUSE_ERROR_TEXT = 'We\'re finalizing your account setup. Please try again in 30 seconds.';
-              if (e.message === ABUSE_ERROR_TEXT &&
-                  requestCount < MAX_REQUESTS) {
+              const ACCOUNT_NOT_YET_FINALIZED_ERROR =
+                  'We\'re finalizing your account setup. Please try again in 30 seconds.';
+              if (e.message === ACCOUNT_NOT_YET_FINALIZED_ERROR && requestCount < MAX_REQUESTS) {
+                // DigitalOcean is still validating this account and may take
+                // up to 30 seconds.  We can retry more frequently to see when
+                // this error goes away.
                 const RETRY_TIMEOUT_MS = 5000;
                 setTimeout(makeRequestRecursive, RETRY_TIMEOUT_MS);
               } else {
--- a/src/server_manager/cloud/digitalocean_api.ts
+++ b/src/server_manager/cloud/digitalocean_api.ts
@@ -124,16 +124,36 @@ class RestApiSession implements DigitalOceanSession {
     // confusing email with their droplet password, which could get mistaken for
     // an invite.
     return this.registerKey_(dropletName, publicKeyForSSH).then((keyId: number) => {
-      return this.request<{droplet: DropletInfo}>('POST', 'droplets', {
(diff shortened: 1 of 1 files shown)


src/server_manager/cloud/digitalocean_api.ts
```diff
@@ -124,16 +124,44 @@ class RestApiSession implements DigitalOceanSession {
...
+              size: dropletSpec.size,
+              image: dropletSpec.image,
+              ssh_keys: [keyId],
+              user_data: dropletSpec.installCommand,
+              tags: dropletSpec.tags,
+              ipv6: true,
+            })
+            .then(fulfill)
+            .catch((e) => {
+              const ACCOUNT_NOT_YET_FINALIZED_ERROR =
```
trevj: This seems brittle. Let's check the `id` or even just the status code, i.e. 403.

src/server_manager/cloud/digitalocean_api.ts
```diff
@@ -124,16 +124,44 @@ class RestApiSession implements DigitalOceanSession {
...
+                  'We\'re finalizing your account setup. Please try again in 30 seconds.';
+              if (e.message === ACCOUNT_NOT_YET_FINALIZED_ERROR && requestCount < MAX_REQUESTS) {
+                // DigitalOcean is still validating this account and may take
+                // up to 30 seconds.  We can retry more frequently to see when
+                // this error goes away.
+                const RETRY_TIMEOUT_MS = 5000;
```
trevj: You could declare this at a higher scope, e.g. near `MAX_REQUESTS`.

src/server_manager/cloud/digitalocean_api.ts
dborkan: I just updated the error message we are actually getting here to be `XHR forbidden failed with 403: We\'re finalizing your account setup. Please try again in 30 seconds.` (the "XHR forbidden failed with 403: " is actually prepended by the `request` method lower in this file).  Does this seem OK?  At this point in the code we don't have the actual XHR error or object, just the error passed to `Promise.reject` from the `request` method.

src/server_manager/cloud/digitalocean_api.ts
dborkan: Done

src/server_manager/cloud/digitalocean_api.ts
trevj: Ah you don't have that field...okay then how about something a little less brittle, e.g. `message.toLower().contains('finalizing')`?

Then, LGTM.

src/server_manager/electron_app/build_action.sh
```diff
@@ -50,4 +50,4 @@ yarn install --prod --ignore-scripts
 
 # Icons.
 cd $ROOT_DIR
-$NODE_MODULES_BIN_DIR/electron-icon-maker --input=src/server_manager/images/launcher-icon.png --output=build/server_manager/electron_app/static
+#$NODE_MODULES_BIN_DIR/electron-icon-maker --input=src/server_manager/images/launcher-icon.png --output=build/server_manager/electron_app/static
```
trevj: Don't forget to uncomment this!

Issue #67, opened 2018-04-04
Long UI loading time of macOS app


testcaoy7:
I just get the Outline client software through Mac App Store and I have find that its UI takes long time to load (at least 3 min).

trevj:
Sorry you've been having trouble with Outline, @testcaoy7. Can you please tell us what version of macOS you're running and roughly the specifications of your system?

testcaoy7:
I am running the latest Outline from App Store and my macOS is the latest High Sierra.
My device is a 2017 model of MacBook Pro with Touch Bar.

Please note that I am using Outline from China.

testcaoy7:
Both Outline server manager and client has the same issue.

jab:
What do you see during the 3 minutes while it’s loading? E.g. Does the app UI appear but with invisible text?

(Reminder: All of Polymer’s paper elements depend on Polymer’s font-roboto element, which tries to download Roboto from fonts.google.com if it’s not locally installed (which is the default case on macOS). And fonts.google.com has been blocked in China. Worse, by default Safari/WKWebView used to paint invisible text while waiting forever for its web font to load, though it recently changed that to 3 seconds.

jab:
(Sorry, the demons in my phone freaked out while I was editing that comment and closed the issue while I was mid-sentence (they must really hate web fonts). Amended and reopened.)

jab:
@alalamav Using Polymer's [`font-roboto`](https://github.com/PolymerElements/font-roboto), which is what happens by default when you include the paper elements, rather than using [`font-roboto-local`](https://github.com/PolymerElements/font-roboto-local), is exactly the problem.

Take a look at font-roboto's html include:
https://github.com/PolymerElements/font-roboto/blob/master/roboto.html#L10

That relies on fonts.googleapis.com being accessible just so it can download the stylesheet that tells it which fonts to use.

Compare that to what font-roboto-local does:
https://github.com/PolymerElements/font-roboto-local/blob/master/roboto.html

That includes the `<style>...</style>` content inline locally, and references the required .ttf files that are included locally too:
https://github.com/PolymerElements/font-roboto-local/tree/master/fonts/roboto

This guarantees that on systems that can't access fonts.googleapis.com, they'll still have all they need locally to properly display `font-family: Roboto` -styled content, without painting it in invisible ink while the fonts.googleapis.com connection times out.

[...]

PR #96, opened 2018-04-13
use font-roboto-local

trevj:
Addresses:
https://github.com/Jigsaw-Code/outline-server/issues/67#issuecomment-381012609

From eyeballing it on my macOS laptop, the app looks identical before and after this change.

### DIFF
--- a/src/server_manager/electron_app/digital_ocean_modifications.ts
+++ b/src/server_manager/electron_app/digital_ocean_modifications.ts
@@ -333,8 +333,9 @@ function addDigitalOceanSignupBanner(text?: string) {
   // we should not rely on any frameworks like Polymer, as they might conflict with
   // existing frameworks on those pages.
   // All class names should begin with Outline to ensure no collisions.
+  // The outline:// protocol should be marked secure to avoid mixed content warnings.
   bannerDiv.innerHTML = `
-      <link href="https://fonts.googleapis.com/css?family=Roboto:300,400,500" rel="stylesheet">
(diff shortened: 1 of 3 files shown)


src/server_manager/electron_app/digital_ocean_modifications.ts
```diff
@@ -333,8 +333,9 @@ function addDigitalOceanSignupBanner(text?: string) {
   // we should not rely on any frameworks like Polymer, as they might conflict with
   // existing frameworks on those pages.
   // All class names should begin with Outline to ensure no collisions.
+  // The outline:// protocol should be marked secure to avoid mixed content warnings.
   bannerDiv.innerHTML = `
-      <link href="https://fonts.googleapis.com/css?family=Roboto:300,400,500" rel="stylesheet">
+      <link href="outline://web_app/bower_components/font-roboto/roboto.html" rel="stylesheet">
```
jab: Now that an html include is being used instead of a CSS stylesheet, I think this should be `rel="import"` and not `rel="stylesheet"`? Maybe the webview does the right thing anyway, but still seems worth changing?

trevj:
@alalamav PTAL, I tested (better) and replaced the logo with an SVG supplied by UX.

src/server_manager/electron_app/digital_ocean_modifications.ts
alalamav: Seems like the change didn't make it through?

trevj:
@alalamav Apologies, all changes now pushed.

PR #120, opened 2018-05-04
prepare for 1.0.6 release

trevj:
Two changes:
* bump version
* remove last vestiges of old auto-update system (followup to https://github.com/Jigsaw-Code/outline-server/pull/103)

Shortly after releasing 1.0.6 we should update `release_data.json` in such a way that it will prompt users to download the new file.


src/server_manager/electron_app/release_scripts/bump_electron_version.js
```diff
@@ -1,39 +0,0 @@
...
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
-
-// Increments the "version" field in the Outline Manager Electron app's
-// config.json file.
-
-const fs = require('fs');
-const semver = require('semver');
```
alalamav: I don't think we use semver elsewhere. We should remove it from our dependencies.

trevj: Good catch. Done.

trevj:
@alalamav That's right regarding `latest.yml`.

PR #143, opened 2018-05-22
 DigitalOcean Agent for Monitoring

### LINKED ISSUE #135: Add DigitalOcean Agent for Monitoring to server
New DigitalOcean Agent for Monitoring is available for monitoring the droplet*. manual installation in not possible because of limited SSH connectivity on outline server.
However it can be added to outline server for all DO/outline users.

* https://www.digitalocean.com/community/tutorials/how-to-install-and-use-the-digitalocean-agent-for-monitoring

cameronaaron:
Fixes #135

Added support DigitalOcean Agent for Monitoring


fortuna:
Thanks for your contribution!

We actually had monitoring enabled in the past, but it made the server creation significantly slower (something like 5 minutes!).
We used to install monitoring via the create droplet API:
https://developers.digitalocean.com/documentation/v2/#create-a-new-droplet
(see `monitoring` field)
The droplet creator would install the monitoring before running our install script.

Do you know how much longer the installation take with your change?
If it takes significantly longer, then maybe we can run it in parallel?

cameronaaron:
This process does take about 4 extra minutes to do so it would be a good solution to run it in parallel with the droplet install. I will need to test to see if that works well though. Thank you for the review

trevj:
@cameronaaron Thanks for contributing!

Here's how we can proceed:
- Let's make the changes in `src/server_manager/install_scripts/do_install_server.sh`. This file gets executed on new droplets.
- Let's use the `curl`-based one liner described here: https://www.digitalocean.com/docs/monitoring/quickstart/#enable-the-digitalocean-agent-on-existing-droplets
- Rather than installing in parallel, we can just call the one-line installer after the `wait` command. This way, the server manager won't block on its completion.

cameronaaron:
That should do it!

src/server_manager/install_scripts/do_install_server.sh
```diff
@@ -174,3 +175,5 @@ done
 # Wait for install script to finish, so that if there is any error in install_server.sh,
 # the finish trap in this file will be able to access its error code.
 wait $install_pid
+curl -sSL https://agent.digitalocean.com/install.sh | sh
```
trevj: Let's add a little commentary here, e.g.:
```sh
wait $install_pid

# Install the DigitalOcean Agent, for improved monitoring:
# https://www.digitalocean.com/docs/monitoring/quickstart/#enable-the-digitalocean-agent-on-existing-droplets
# 
# Since the server manager looks only for the tags created in the previous
# step, this does not slow down server creation.
curl -sSL https://agent.digitalocean.com/install.sh | sh
```

cameronaaron:
Github has been giving me issues today but I think this should be correct now hopefully

src/server_manager/install_scripts/do_install_server.sh
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env bash
```
trevj: let's remove this for now

scripts/do_action.sh
```diff
@@ -35,3 +35,5 @@ function do_action() {
 export -f do_action
 
 do_action "$@"
+
```
trevj: again, let's revert

src/server_manager/install_scripts/do_install_server.sh
```diff
@@ -174,3 +175,9 @@ done
 # Wait for install script to finish, so that if there is any error in install_server.sh,
 # the finish trap in this file will be able to access its error code.
 wait $install_pid
+# Install the DigitalOcean Agent, for improved monitoring:
```
trevj: how about a blank line here?

cameronaaron:
There we go :)

cameronaaron:
very sorry I am still getting used to using GIT so this is really good practice. I am sorry that I am being a bit slow. I do appreciate your understanding. I believe the fixes are all in now for this pull after this I will fix the next pull. I think I made the mistake of trying to improve too many things at once and now there is too much to revert and fix. Please let me know if there is anything else needed for this

trevj:
@cameronaaron Thanks for sticking with it - it's not easy to get started with git! There is still one change to `app.ts` but it should be there, I'm going to go ahead and squash merge this.

cameronaaron:
Thank you so much! I am glad!

cameronaaron:
Now off to the next one haha

trevj:
Hmm, something weird with the Travis/GitHub integration; will try closing and then re-opening this to trigger it.

fortuna:
Thanks @cameronaaron !!! The extra monitoring will help us better understanding the resource consumption of Outline servers!

cameronaaron:
Ofcourse!! I hope to get some more improvements implemented in the future! :) Thank you for the opportunity. I am a CS student so any opportunities to contribute to code is invaluable

fortuna:
By the way, with a successful installation, the droplet graphs should look like this: 
image

PR #155, opened 2018-05-29
Replace safe_docker with dynamic DOCKER_CMD

alalamav:
Fixes firewall detection when the user is root, as the `safe_docker` function would have not been defined.


alalamav:
I don't fully recall, but there was a good reason, related to the DO environment, for using this scheme. @trevj will remember the details. Can we merge this PR as is to fix the immediate issue, and implement your suggestion in a follow up PR?

PR #202, opened 2018-07-12
Clean up shell variable expansion

cameronaaron:


cameronaaron:
Didn't mean to open a second pull

PR #203, opened 2018-07-12
Add shebangs to bash files 

cameronaaron:
Added she bangs to bash files

PR #204, opened 2018-07-12
CSS clean up-- Remove unit on zero value

cameronaaron:
This is just a suggestion but for many devs 0 without units is easier to read and it will also cut down on the css a bit.


trevj:
So, as for https://github.com/Jigsaw-Code/outline-server/pull/203, this would be easier to review if you can branch from master - the resulting diff will be so much smaller.

trevj:
Closing for now - see https://github.com/Jigsaw-Code/outline-server/pull/205.

PR #237, opened 2018-08-24
migrate to Electron 2.x

trevj:
Magic mode and advanced modes seem to work fine with Electron 2; it complained about our content security policy which turned out to be completely bogus so I updated it.

BTW, I tried electron-updater 3.x but ran into compile errors. I don't think it's ready for primetime yet.

### DIFF
--- a/src/server_manager/index.html
+++ b/src/server_manager/index.html
@@ -15,45 +15,40 @@
 -->
 <!DOCTYPE html>
 <html>
-  <head>
-    <meta charset="UTF-8">
-    <!--<base href=".">-->
-    <meta name="google" content="notranslate">
-    <meta http-equiv="Content-Language" content="en">
-    <meta name="viewport" content="width=device-width, user-scalable=no, initial-scale=1.0, minimum-scale=1.0, maximum-scale=1.0">
-    <!-- TODO: remove this once launched to a broader audience -->
-    <meta name="robots" content="noindex,noarchive">
-    <!-- TODO: update this description -->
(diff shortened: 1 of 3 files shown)


src/server_manager/index.html
```diff
@@ -15,45 +15,40 @@
...
+
+  <link rel="shortcut icon" href="/ui_components/images/favicon.png">
+  <link rel="stylesheet" href="/ui_components/style.css">
+  <script src="/bower_components/webcomponentsjs/webcomponents-loader.js"></script>
+  <link rel="import" href="/bower_components/app-layout/app-toolbar/app-toolbar.html">
+  <link rel="import" href="/ui_components/app-root.html">
+  <script src="server_manager/web_app/main.js"></script>
+
+  <!-- TODO: Add Outline servers to connect-src on-demand (in addition to the DigitalOcean API). -->
+  <meta http-equiv="Content-Security-Policy" content="default-src 'unsafe-inline' outline:; connect-src https:; frame-src https://s3.amazonaws.com/outline-vpn/">
```
alalamav: 👍what about the other non-http protocols, i.e. macappstore://, ss://?

alalamav: Are you sure? The invitation page links to `macappstore://` for downloading the macOS client and `ss://` for importing the server when the app is already installed.

trevj: *click* The iframe-d invite? I just tested and:
- Mac App Store opens without mapappstore:// in the CSP
- ss:// link doesn't open with or without macappstore:// in the CSP (in other words, it's broken as-is - I just filed an issue)

PR #240, opened 2018-08-28
Add ss: protocol to frame-src CSP

alalamav:
Fixes opening `ss://` links in the "get connected" flow.


trevj:
Oh nice, thank you.

PR #259, opened 2018-10-01
enable BBR on new DigitalOcean droplets

trevj:
Tested this just now with a DigitalOcean server in London, seems to work great.


trevj:
@sandrigo FYI this is for the magic experience

sandrigo:
Mega W00t!

PR #268, opened 2018-10-10
Restrict Shadowbox outbound access to localhost

alalamav:
* Uses the `--acl` option in `ss-server` to restrict outbound access to local and multicast subnets.
* This should be removed once we migrate to shadowsocks-go.

FYI, @fortuna.


alalamav:
Sure! Done in 51c006c.

PR #273, opened 2018-10-16
Re-style manager UI

alalamav:
* Refactors the manager app UI by updating the layout and color scheme.
* Adds responsiveness to the display, allowing the window to be resized.
* Implements a collapsable drawer.
* Serves as a building block for supporting multiple servers.

screen shot 2018-10-16 at 12 26 29 pm
screen shot 2018-10-16 at 12 28 01 pm
screen shot 2018-10-16 at 12 25 13 pm
screen shot 2018-10-16 at 12 24 48 pm
screen shot 2018-10-16 at 12 24 28 pm
screen shot 2018-10-16 at 12 24 04 pm


src/server_manager/electron_app/index.ts
```diff
@@ -70,9 +70,10 @@ interface IpcEvent {
 
 function createMainWindow() {
   const win = new electron.BrowserWindow({
-    width: 600,
-    height: 768,
-    resizable: false,
+    width: 800,
```
trevj: :tada:

src/server_manager/ui_components/app-root.html
```diff
@@ -42,8 +43,20 @@
     <style>
       :host {
```
trevj: It doesn't look like `width: 100%;` is needed anymore (perhaps it never was)?

src/server_manager/ui_components/app-root.html
```diff
@@ -42,8 +43,20 @@
...
         width: 100%;
-        --primary-green: #00bfa5;
-        --light-gray: color: rgba(0,0,0,0.54);
+        --status-bar-width: 48px;
```
trevj: +1 for variable names. I think we can improve this one - throughout this file I see several names for the sidebar and it changes when it's minimised/maximised. I think we should settle on names for the bar and its two states - it could be "sidebar" along with "narrow" and "wide" (or empty string, to indicate that that's the normal state).

WDYT?

trevj: Also, how did you choose this figure? If I set it to something obviously crazy, e.g. 200px, then I get a nasty jump when minimising/maximising.

src/server_manager/ui_components/app-root.html
```diff
@@ -42,8 +43,20 @@
...
+      }
+      .app-container {
+        margin: 0 auto;
+      }
+      @media (min-width: 1281px) {
```
trevj: There's an implied magic figure of 80px in these `@media` declarations - how did you arrive at that figure?

src/server_manager/ui_components/app-root.html
```diff
@@ -59,126 +72,178 @@
...
       }
       app-toolbar [main-title] img {
         height: 16px;
         margin-top: 8px;
       }
+      .side-bar-margin {
+        margin-left: var(--status-bar-width);
+      }
+      #sideBar {
+        --app-drawer-width: var(--status-bar-width);
```
trevj: This variable doesn't seem to be used anywhere else?

src/server_manager/ui_components/app-root.html
```diff
@@ -59,126 +72,178 @@
...
+        --app-drawer-content-container: {
+          background-color: var(--background-contrast-color);
+        };
+      }
+      #sideBar div {
+        margin-top: 12px;
+        height: 100%;
```
trevj: It doesn't look like this is necessary?

src/server_manager/ui_components/cloud-install-styles.html
```diff
@@ -206,6 +214,23 @@
       [hidden] {
         display: none !important;
       }
+
+      .flex-1 {
```
trevj: Where are these used?

src/server_manager/ui_components/outline-do-oauth-step.html
```diff
@@ -31,84 +31,94 @@
...
-              <img src="images/digital_ocean_logo.svg" />
-              <p>Waiting to connect your account...</p>
-            </div>
-            <paper-button on-tap="_cancelTapped">Cancel</paper-button>
-          </paper-card>
-        </outline-step-view>
+    <iron-pages id='pages' attr-for-selected='id' selected='{{ currentPage }}'>
+      <outline-step-view id="connectAccount">
+        <span slot="step-title">Sign in or create an account with DigitalOcean.</span>
+        <span slot="step-description">With your account, Outline makes it easy to<br/> create a server and get connected.</span>
```
trevj: Oh gosh, `<br/>`? Let's remove these now if it's easy - otherwise, file an issue.

src/server_manager/ui_components/app-root.html
alalamav: Polymer uses to set the width of the drawer.

src/server_manager/ui_components/app-root.html
alalamav: This will be necessary to ensure that the elements within the side bar occupy the its full height.

src/server_manager/ui_components/app-root.html
alalamav: These are the "standard" resolutions for tablet, laptop, and desktop. I have added comments.
(See these for [more](https://gist.github.com/gokulkrishh/242e68d1ee94ad05f488) [information](https://css-tricks.com/snippets/css/media-queries-for-standard-devices/)).

src/server_manager/ui_components/app-root.html
alalamav: The width is defined in the mocks. I think the animation works at the current width, the effect you describe could be the interaction of the side bar and app drawer animations.

trevj: Yep, that makes sense though I found it a little surprising - I would have assumed the drawer widget itself supported a minimised state.

trevj:
Looking good. For bonus points, I see a little issue when expanding/collapsing the app drawer: the background "jumps" noticeably, as the amount of horizontal space changes. There may be an easy fix.

PR #284, opened 2018-10-23
Update from alpha-series Cloud Functions to beta

bemasc:
This is a significant change because the beta branch explicitly
supports only Node 6.x, and will fail during setup if you are
using another node version.  This is unfortunate, but it's
better than our current behavior where "yarn" fails with
a compilation error on recent Node versions.


trevj:
Closing now that https://github.com/Jigsaw-Code/outline-server/pull/347 is merged.

PR #288, opened 2018-10-25
Documentation update

sandrigo:
Updating documentation given user feedback, clarifying where to obtain the appropriate parameters to issue an API call to the server. This also simplifies construction of the URL and allows for calling the server without needing to be SSH'd in in the first place.

### DIFF
--- a/src/shadowbox/README.md
+++ b/src/shadowbox/README.md
@@ -81,26 +81,30 @@ docker rmi $(docker images -f dangling=true -q)
 
 ## Access Keys Management API
 
+In order to Access the Management API, you'll need to know the apiUrl for your server. You can obtain this information from the access.txt file under the 'shadowbox' directory in your Outline server. An example apiUrl is: https://1.2.3.4:1234/3pQ4jf6qSr5WVeMO0XOo4z. 
+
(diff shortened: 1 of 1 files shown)


src/shadowbox/README.md
```diff
@@ -81,26 +81,30 @@ docker rmi $(docker images -f dangling=true -q)
...
 ## Access Keys Management API
 
+In order to utilize the Management API, you'll need to know the apiUrl for your Outline server. You can obtain this information from the 'access.txt' file under the 'shadowbox' directory of your server. An example apiUrl is: https://1.2.3.4:1234/3pQ4jf6qSr5WVeMO0XOo4z. 
+
+By replacing apiUrl with the actual URL you see in that file, you can programmatically perform the following operations on the server.
+
 List users
 ```
-curl --insecure https://localhost:8081/TestApiPrefix/access-keys/
+curl --insecure apiUrl/access-keys/
```
alalamav: You could replace these with a bash variable: `$API_URL` (in the example, `API_URL=https://1.2.3.4:1234/3pQ4jf6qSr5WVeMO0XOo4z`) That way, it will be easier to define the variable in a terminal and copy/paste the example commands without having to edit them.

PR #293, opened 2018-10-26
Sort keys by name

DavidLemayian:
Currently keys are sorted by creation dates.

When we have many users, it makes it easy manage keys if it's sorted by name. E.g when you have multiple orgs/departments on one server, you can have:
```
cfa-david
...
alpha-larry
...
jigsaw-george
```

Above by sorting by name would know where to scroll to to find it without introducing any new "features" e.g search nor trying to remember when you created the key:

```
alpha-larry
...
cfa-david
...
jigsaw-george
```

### DIFF
--- a/src/server_manager/ui_components/outline-server-view.html
+++ b/src/server_manager/ui_components/outline-server-view.html
@@ -330,7 +330,7 @@
         <paper-button class='connectButton' id='getConnectedText' on-tap='_handleConnectPressed'>GET CONNECTED</paper-button>
       </div>
       <!-- rows for each access key -->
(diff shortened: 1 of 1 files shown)


DavidLemayian:
I signed it!

DavidLemayian:
Sure thing @alalamav. I'll check out branch from that cause I notice it doesn't yet have master branch changes.

alalamav:
That's exactly what I was thinking. The sort attribute of dom-repeat supports a function that returns a sort function; therefore, we can change it depending on the selected sorting property.

DavidLemayian:
Cool @alalamav. Sounds interesting enough to take a stab at it. I'll make some time over the weekend unless you were planning to work on it sooner?

PR #305, opened 2018-11-05
docker image for ARM architecture.

aguilajesus:
Dockerfile to build image for raspberry Pi, it won't work with the current prometheus binaries, do you want me to add the binaries to the PR or try to find a way to build them on the fly (or even to pull them from other docker image)


shaomeng:
Is this PR taken care of by the Outline team? I'm very interested in deploying Outline on the raspberry pi I have at home, and would love to offer test if needed.

Issue #78, opened 2018-04-07
[Manager] Multiple Server Support.


laosb:
Well done!

We're currently using this across many departments, and we surely needs more than just one server to proxy all connections for our team members (more than 1000+ long connections to one infamous sounds like easy to be found and block). Could you support setting up more than one server per Manager instance?

PS: I'm also thinking should we have some ways to backup server managing credentials? Maybe another issue to open.

ow:
This is the first thing I wanted, and the thing I came here to look for. Adding my support for this, as having a single region is quite frustration/a blocker for me!

trevj:
@laosb and @ow, thanks for the feedback and suggestions!

tsekityam:
I can manage multiple servers with Outline Manager by backup/restore the local storage (i.e. `~/Library/Application Support/Outline Manager/Local Storage/` for macOS) of the manager. 
It is not a good solution but, at least, it works.

Nz666-Kshatriya:
The same suggestion as I'd like to give

MaXFalstein:
I would like to take this further by supporting DigitalOcean load balancers and multiple droplets.

trevj:
Popular feature!

Duplicates:
- https://github.com/Jigsaw-Code/outline-server/issues/127
- https://github.com/Jigsaw-Code/outline-server/issues/112
- https://github.com/Jigsaw-Code/outline-server/issues/122

viralking:
Agree! This would be awesome, to have multiple locations in the manager :)

xukaifu:
Awesome, I'm looking forward to it!

NEXUS2345:
In terms of multiple servers, could we allow multiple servers from different providers? The ability to diversify the providers used would make the service more stable and provide more options for people who are getting some providers blocked in some countries but not others.

trevj:
@alalamav Can you add a little colour here to what you've been working on? Thanks.

alalamav:
This feature is indeed prioritized. I have been working on a UI refactor (#273) that will enable us to add support for managing multiple servers. Here's a sneak peek into our team's vision for the Outline manager app:

![mss-mock](https://user-images.githubusercontent.com/2132122/47032902-7de46680-d141-11e8-99c0-ec0064d6d39a.png)

NEXUS2345:
@alalamav this looks really nice. I'm looking forward to getting to use it. Thank you very much for your work.

alalamav:
[...]

PR #306, opened 2018-11-07
Enable multiple server support

alalamav:
* Enables the management of multiple servers.
* TODO
  * App tests.
  * DigitalOcean offline mode.

### DIFF
--- a/src/server_manager/web_app/app.ts
+++ b/src/server_manager/web_app/app.ts
@@ -236,7 +236,7 @@ export class App {
 
   private async syncServersToDisplay(servers: server.Server[]) {
     for (const server of servers) {
-      this.syncServerToDisplay(server);
+      await this.syncServerToDisplay(server);
     }
   }
 
@@ -261,7 +261,7 @@ export class App {
       this.syncDisplayServersToUi();
     } else {
       // We may need to update the stored display server if it was persisted when the server was not
-      // healthy.
+      // healthy, or the server has been renamed.
       try {
         const remoteServerName = server.getName();
         if (displayServer.name !== remoteServerName) {
@@ -471,7 +471,7 @@ export class App {
         });
       } else {
         // Display the unreachable server state within the server view.
-        const serverView = this.appRoot.getServerView();
+        const serverView = this.appRoot.getServerView(displayServer.id);
         serverView.isServerReachable = false;
         serverView.isServerManaged = isManagedServer(server);
         serverView.serverName = displayServer.name;  // Don't get the name from the remote server.
@@ -654,7 +654,7 @@ export class App {
           const server = this.serverBeingCreated;
           this.serverBeingCreated = null;
           this.syncServerToDisplay(server).then((displayServer) => {
-            this.showServer(server, displayServer);
+            this.showServerAfterTimeout(server, displayServer);
           });
         })
         .catch((e) => {
@@ -704,6 +704,15 @@ export class App {
         });
   }
 
+  // Shows a server after a `timeoutMs`. This is sometimes necessary to guarantee that the UI gets
+  // updated before displaying a server.
+  private showServerAfterTimeout(
+      selectedServer: server.Server, selectedDisplayServer: DisplayServer, timeoutMs = 250) {
+    setTimeout(() => {
+      this.showServer(selectedServer, selectedDisplayServer);
+    }, timeoutMs);
+  }
+
   // Show the server management screen. Assumes the server is healthy.
   private showServer(selectedServer: server.Server, selectedDisplayServer: DisplayServer): void {
     this.selectedServer = selectedServer;
@@ -711,7 +720,7 @@ export class App {
     this.displayServerRepository.storeLastDisplayedServerId(selectedDisplayServer.id);
 
     // Show view and initialize fields from selectedServer.
-    const view = this.appRoot.getServerView();
+    const view = this.appRoot.getServerView(selectedDisplayServer.id);
     view.isServerReachable = true;
     view.serverId = selectedServer.getServerId();
     view.serverName = selectedServer.getName();
@@ -729,14 +738,6 @@ export class App {
       view.serverLocation = digitalocean_server.GetEnglishCityName(host.getRegionId());
     } else {
       view.isServerManaged = false;
-      // TODO(dborkan): consider using dom-if with restamp property
-      // https://www.polymer-project.org/1.0/docs/api/elements/dom-if
-      // or using template-repeat.  Then we won't have to worry about clearing
-      // the server-view when we display a new server.  This should be fixed
-      // once we support multiple servers.
-      view.serverLocation = undefined;
-      view.monthlyCost = undefined;
-      view.monthlyOutboundTransferBytes = undefined;
     }
 
     view.metricsEnabled = selectedServer.getMetricsEnabled();
@@ -843,7 +844,7 @@ export class App {
     this.selectedServer.addAccessKey()
         .then((serverAccessKey: server.AccessKey) => {
           const uiAccessKey = convertToUiAccessKey(serverAccessKey);
-          this.appRoot.getServerView().addAccessKey(uiAccessKey);
+          this.appRoot.getServerView(this.appRoot.selectedServer.id).addAccessKey(uiAccessKey);
           this.displayNotification('Key added');
         })
         .catch((error) => {
@@ -900,7 +901,7 @@ export class App {
       return manualServer.isHealthy().then((isHealthy) => {
         if (isHealthy) {
           return this.syncServerToDisplay(manualServer).then((displayServer) => {
-            this.showServer(manualServer, displayServer);
+            this.showServerAfterTimeout(manualServer, displayServer);
           });
         } else {
           // Remove inaccessible manual server from local storage if it was just created.
@@ -916,7 +917,7 @@ export class App {
   private removeAccessKey(accessKeyId: string) {
     this.selectedServer.removeAccessKey(accessKeyId)
         .then(() => {
-          this.appRoot.getServerView().removeAccessKey(accessKeyId);
+          this.appRoot.getServerView(this.appRoot.selectedServer.id).removeAccessKey(accessKeyId);
           this.displayNotification('Key removed');
         })
         .catch((error) => {
@@ -982,7 +983,8 @@ export class App {
     this.selectedServer.setMetricsEnabled(metricsEnabled)
         .then(() => {
           // Change metricsEnabled property on polymer element to update display.
-          this.appRoot.getServerView().metricsEnabled = metricsEnabled;
+          this.appRoot.getServerView(this.appRoot.selectedServer.id).metricsEnabled =
+              metricsEnabled;
         })
         .catch((error) => {
           this.displayError('Error setting metrics enabled', error);
@@ -992,10 +994,9 @@ export class App {
   private renameServer(newName: string): void {
     this.selectedServer.setName(newName)
         .then(() => {
-          this.appRoot.getServerView().serverName = newName;
-          this.removeServerFromDisplay(this.appRoot.selectedServer);
+          this.appRoot.getServerView(this.appRoot.selectedServer.id).serverName = newName;
           this.syncServerToDisplay(this.selectedServer).then((displayServer) => {
-            this.appRoot.selectedServer = displayServer;
+            this.showServerAfterTimeout(this.selectedServer, displayServer);
           });
         })
         .catch((error) => {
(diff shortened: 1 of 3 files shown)


src/server_manager/web_app/app.ts
```diff
@@ -200,136 +209,218 @@ export class App {
...
 
   start(): void {
-    // Load manual servers from storage.
-    this.manualServerRepository.listServers().then((manualServers) => {
-      // Show any manual servers if they exist.
-      if (manualServers.length > 0) {
-        this.showManualServerIfHealthy(manualServers[0]);
-        return;
-      }
+    // Load display servers and associate them with managed and manual servers.
```
trevj: I think we can make this function easier to understand.

Seems like what really happens is:
- fetch the last-known state of servers from localStorage
- send that to the UX
- launch a bunch of queries to update the state of each server

Let's update the comment to make that clearer.

I think we can also:
- reduce the level of indentation
- de-dupe the command a bit

e.g.:
```ts
  async start() {
    this.appRoot.serverList = await this.displayServerRepository.listServers();

    const manualServers = this.manualServerRepository.listServers();

    const accessToken = this.digitalOceanTokenManager.getStoredToken();
    const managedServers =
        accessToken ? this.enterDigitalOceanMode(accessToken) : Promise.resolve([]);

    Promise.all([manualServers, managedServers])
        .then(([a, b]) => {
          return this.syncServersToDisplay(a.concat(b));
        })
        .then(() => {
          this.maybeShowLastDisplayedServer();
        });

    this.showIntro();
  }
```

src/server_manager/web_app/app.ts
```diff
@@ -200,136 +209,218 @@ export class App {
...
       }
+    });
+    this.showIntro();
+  }
 
-      // User has no manual servers or DigitalOcean token.
-      this.showIntro();
+  private async syncServersToDisplay(servers: server.Server[]) {
+    for (const server of servers) {
+      await this.syncServerToDisplay(server);
```
trevj: Why wait - won't this only slow things down?

src/server_manager/web_app/app.ts
```diff
@@ -200,136 +209,218 @@ export class App {
...
+    }
+  }
+
+  // Syncs the locally persisted server metadata for `server`. Creates a DisplayServer for `server`
+  // if one is not found in storage.
+  private async syncServerToDisplay(server: server.Server): Promise<DisplayServer> {
```
trevj: I think this function tries to do too much. Seems like we need functions to:
- query a server and return a `DisplayServer`
- update the display table
- add new items to the display table

Breaking those functions out will make things more readable overall and help reduce some of the duplicate code currently in this function, viz. multiple calls to `syncDisplayServersToUi``.

src/server_manager/web_app/app.ts
```diff
@@ -200,136 +209,218 @@ export class App {
...
+    const displayServerId = server.getManagementApiUrl();
+    let displayServer = this.displayServerRepository.findServer(displayServerId);
+    if (!displayServer) {
+      console.warn(`Could not find managed display server with ID ${displayServerId}`);
```
trevj: This seems like a normal thing to happen, no?

src/server_manager/web_app/app.ts
```diff
@@ -200,136 +209,218 @@ export class App {
...
+        }
+      } catch (e) {
+        // Ignore, we may not have the server config yet.
+      }
+    }
+    this.serverDisplayTable[displayServer.id] = {displayServer, server};
+    return displayServer;
+  }
+
+  private syncDisplayServersToUi() {
```
trevj: Why don't you call this in `start`?

trevj: This function needs a comment, especially the part about triggering a Polymer mutation - does Polymer not detect changes in the object? Right now you call it every time you fetch a server's info which seems wasteful, albeit reliable.

src/server_manager/web_app/app.ts
```diff
@@ -79,13 +85,16 @@ type DigitalOceanServerRepositoryFactory = (session: digitalocean_api.DigitalOce
 export class App {
   private digitalOceanRepository: server.ManagedServerRepository;
   private selectedServer: server.Server;
-  private runningServer: server.Server;
+  private serverBeingCreated: server.ManagedServer;
```
trevj: Having both is a smell - do you really need them?

src/server_manager/web_app/app.ts
```diff
@@ -79,13 +85,16 @@ type DigitalOceanServerRepositoryFactory = (session: digitalocean_api.DigitalOce
...
+  private displayServerBeingCreated: DisplayServer;
+  private serverDisplayTable: ServerDisplayTable = {};
```
trevj: I'm not convinced this helps us much and there is a bunch of code required to manage it. How would you implement this class without this function? Would it improve the class?

src/server_manager/web_app/app.ts
```diff
@@ -200,136 +209,218 @@ export class App {
...
+
+  // Syncs the locally persisted server metadata for `server`. Creates a DisplayServer for `server`
+  // if one is not found in storage.
+  private async syncServerToDisplay(server: server.Server): Promise<DisplayServer> {
+    const displayServerId = server.getManagementApiUrl();
+    let displayServer = this.displayServerRepository.findServer(displayServerId);
+    if (!displayServer) {
+      console.warn(`Could not find managed display server with ID ${displayServerId}`);
+      const isHealthy = await server.isHealthy().catch((e) => false);
+      displayServer = {
```
trevj: I find avoiding the need to declare a variable often improves readability - it's certainly possible here.

src/server_manager/web_app/app.ts
```diff
@@ -200,136 +209,218 @@ export class App {
...
+  }
+
+  // Shows the last server displayed, if there is one in local storage.
+  private maybeShowLastDisplayedServer() {
+    const lastDisplayedServerId = this.displayServerRepository.getLastDisplayedServerId();
+    if (!lastDisplayedServerId) {
+      return;  // No server was displayed when user quit the app.
+    }
+    const lastDisplayedServer = this.displayServerRepository.findServer(lastDisplayedServerId);
+    if (!lastDisplayedServerId) {
```
trevj: Did you mean to check `lastDisplayedServer` here instead?

src/server_manager/web_app/app.ts
```diff
@@ -200,136 +209,218 @@ export class App {
...
   }
 
+  // Removes `displayServer` from the UI.
+  private removeServerFromDisplay(displayServer: DisplayServer) {
+    delete this.serverDisplayTable[displayServer.id];
+    this.displayServerRepository.removeServer(displayServer);
+    this.syncDisplayServersToUi();
+  }
+
+  // Shows the last server displayed, if there is one in local storage.
```
trevj: "... and it still exists"?

src/server_manager/web_app/app.ts
```diff
@@ -386,6 +477,19 @@ export class App {
   }
 
   private connectToDigitalOcean() {
+    const accessToken = this.digitalOceanTokenManager.getStoredToken();
+    if (accessToken) {
```
trevj: Seems like this whole block is basically duplicated in `start` - and then later in this same function!

If it helps, I noticed this while looking at all the calls to `syncServersToDisplay` - seems like there should only be 2-3 places where that's called, right?

src/server_manager/web_app/display_server.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  storeLastDisplayedServerId(serverId: string) {
+    this.storage.setItem(DisplayServerRepository.LAST_DISPLAYED_SERVER_STORAGE_KEY, serverId);
+  }
+
+  getLastDisplayedServerId(): string {
+    return this.storage.getItem(DisplayServerRepository.LAST_DISPLAYED_SERVER_STORAGE_KEY);
+  }
+
+  removeLastDisplayedSeverId() {
```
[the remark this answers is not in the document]
alalamav: Done in 8b3ab47

src/server_manager/web_app/display_server.ts
```diff
@@ -0,0 +1,88 @@
...
+
+import * as server from '../model/server';
+
+export interface DisplayServer {
+  id: string;
+  name: string;
+  isManaged: boolean;
+}
+
+// Persistence layer for server display data.
```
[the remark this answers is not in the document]
alalamav: Done in 8b3ab47

src/server_manager/web_app/display_server.ts
```diff
@@ -0,0 +1,88 @@
...
+    this.servers = this.servers.filter((server: DisplayServer) => {
+      return server.id !== serverToForget.id;
+    });
+    this.storeServers();
+  }
+
+  findServer(serverId: string): DisplayServer {
+    return this.servers.find(server => server.id === serverId);
+  }
+
```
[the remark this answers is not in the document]
alalamav: Done in 8b3ab47.

src/server_manager/web_app/app.ts
alalamav: Yes, good catch! Done in 8b3ab47

src/server_manager/web_app/app.ts
alalamav: Done in 8b3ab47

src/server_manager/web_app/display_server.ts
```diff
@@ -0,0 +1,88 @@
...
+      try {
+        this.servers = JSON.parse(serversJson);
+      } catch (e) {
+        console.error('Error loading local servers from storage');
+      }
+    }
+  }
+
+  private storeServers() {
+    this.storage.setItem(DisplayServerRepository.SERVERS_STORAGE_KEY, JSON.stringify(this.servers));
```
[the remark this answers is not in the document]
alalamav: According to the [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#Exceptions), it only throws a TypeError if there is a circular reference in the object.

src/server_manager/web_app/app.ts
alalamav: As discussed offline, in this case we have to assign the variable because we use it after the conditional logic.

src/server_manager/web_app/display_server.ts
```diff
@@ -0,0 +1,88 @@
...
+    this.storage.removeItem(DisplayServerRepository.LAST_DISPLAYED_SERVER_STORAGE_KEY);
+  }
+
+  private loadServers() {
+    const serversJson = this.storage.getItem(DisplayServerRepository.SERVERS_STORAGE_KEY);
+    if (serversJson) {
+      try {
+        this.servers = JSON.parse(serversJson);
+      } catch (e) {
+        console.error('Error loading local servers from storage');
```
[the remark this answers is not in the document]
alalamav: As discussed offline, this method is called only at the constructor and `main` creates the repository before the app starts. In this particular case, we may be better off not throwing the exception, as the side-effects of the failure are recoverable.

src/server_manager/web_app/app.ts
alalamav: I have removed the table in fb762f6 and now query the repositories. I think it makes the code more maintainable as you pointed out.

src/server_manager/web_app/app.ts
alalamav: I tried to refactor it into a separate method, but if you notice closely, the logic is slightly different on every call. Please let me know if you have ideas on how to improve this.

trevj:
Related issue:
https://github.com/Jigsaw-Code/outline-server/issues/78

src/server_manager/web_app/app.ts
alalamav: Actually, the wait is needed to ensure that the UI server list is synced after calling this method.

src/server_manager/web_app/app.ts
trevj: This seems like something to tackle in the future.

src/server_manager/web_app/app.ts
trevj: OK. That's worth a comment. Seems like this function could use `Promise.all` to run the queries in parallel. But not a big deal.

src/server_manager/web_app/app.ts
trevj: Excellent.

src/server_manager/web_app/app.ts
alalamav: What I don't like about `Promise.all` is that if a single promise rejects, it rejects. What I have learned from using async/await is that await should always be called to ensure deterministic results. I'll merge now and add the comment in a follow up PR, if that's okay with you.

PR #313, opened 2018-11-16
Handle managed server deletion outside the application

alalamav:
* Displays a 'syncing' icon until cached displayed servers have been synced with their remote objects.
* Removes unsynced managed servers on start, or upon retrying to contact an unreachable server.
* This is only relevant to managed servers, as deletion may occur outside the manager app.


alalamav:
That's right. Prior to this change we were removing the deleted servers only when the user clicked on them.

PR #314, opened 2018-11-19
Re-implement broken tests, write tests for new functionality

alalamav:
This is still far from the coverage that we want, but it matches and extends the previous unit tests.

### DIFF
--- a/src/server_manager/web_app/app.spec.ts
+++ b/src/server_manager/web_app/app.spec.ts
@@ -19,7 +19,7 @@ import * as server from '../model/server';
 
 import {App} from './app';
 import {TokenManager} from './digitalocean_oauth';
-import {DisplayServer, DisplayServerRepository} from './display_server';
+import {DisplayServer, DisplayServerRepository, makeDisplayServer} from './display_server';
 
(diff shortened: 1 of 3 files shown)


src/server_manager/web_app/app.spec.ts
```diff
@@ -64,13 +233,24 @@ function createTestApp(
...
+  }
   return new App(
       polymerAppRoot, WEB_APP_URL, VERSION, fakeDigitalOceanSessionFactory,
-      fakeDigitalOceanServerRepositoryFactory, manualServerRepo, fakeDisplayServerRepository,
+      fakeDigitalOceanServerRepositoryFactory, manualServerRepo, displayServerRepository,
       digitalOceanTokenManager);
 }
 
+async function makeDisplayServer(server: server.Server) {
+  return {
```
trevj: This is a near duplicate of code in `app.ts`. I wonder if it's worth factoring that out to a function. It maybe best-placed in `display}_server.ts`.

alalamav: Good catch, I've implemented your suggestion in f0f930b.

PR #323, opened 2018-11-26
Manual server entry UI improvements

alalamav:
* Hides the underline in the manual server input textarea.
* Defines a mixin to reduce style duplication.


trevj:
Followup to:
https://github.com/Jigsaw-Code/outline-server/pull/315

PR #333, opened 2018-12-04
Multiple server support

alalamav:
Merging this set of approved PRs

FYI @fortuna, @trevj.


trevj:
Just to note that myself and @alalamav manually verified all required CLAs were signed.

PR #349, opened 2018-12-13
Implement cloud function to post Sentry events to Salesforce

alalamav:
* The cloud function and enclosing package follow the example of `metrics_server`.
* It receives Sentry events via webhook integration, extracts the data and posts it as form data to Salesforce.
* Infers whether the event is from development or production and posts to the corresponding Salesforce endpoint.

### DIFF
--- a/src/sentry_webhook/index.ts
+++ b/src/sentry_webhook/index.ts
@@ -1,0 +1,47 @@
+// Copyright 2018 The Outline Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as sentry from '@sentry/types';
+import * as express from 'express';
+
+import {postSentryEventToSalesforce} from './post_sentry_event_to_salesforce';
+
+exports.postSentryEventToSalesforce = (req: express.Request, res: express.Response) => {
+  if (req.method !== 'POST') {
+    res.status(405).send('Method not allowed');
+    return;
+  }
+  if (!req.body) {
+    res.status(400).send('Missing request body');
+    return;
+  }
+
+  const sentryEvent: sentry.SentryEvent = req.body.event;
+  if (!sentryEvent) {
+    res.status(400).send('Missing Sentry event');
+    return;
+  }
+  // Use the request message if SentryEvent.message is unpopulated.
+  sentryEvent.message = sentryEvent.message || req.body.message;
+  postSentryEventToSalesforce(sentryEvent, req.body.project)
+      .then(() => {
+        res.status(200).send();
+      })
+      .catch((e) => {
+        console.error(e);
+        // Send an OK response to Sentry - they don't need to know about errors with posting to
+        // Salesforce.
+        res.status(200).send();
+      });
+};
--- a/src/sentry_webhook/index.ts
+++ b/src/sentry_webhook/index.ts
@@ -15,22 +15,22 @@
 import * as sentry from '@sentry/types';
 import * as express from 'express';
 
-import {postSentryEventToSalesforce} from './post_sentry_event_to_salesforce';
+import {postSentryEventToSalesforce, shouldPostEventToSalesforce} from './post_sentry_event_to_salesforce';
 
 exports.postSentryEventToSalesforce = (req: express.Request, res: express.Response) => {
   if (req.method !== 'POST') {
-    res.status(405).send('Method not allowed');
-    return;
+    return res.status(405).send('Method not allowed');
   }
   if (!req.body) {
-    res.status(400).send('Missing request body');
-    return;
+    return res.status(400).send('Missing request body');
   }
 
   const sentryEvent: sentry.SentryEvent = req.body.event;
   if (!sentryEvent) {
-    res.status(400).send('Missing Sentry event');
-    return;
+    return res.status(400).send('Missing Sentry event');
+  }
+  if (!shouldPostEventToSalesforce(sentryEvent)) {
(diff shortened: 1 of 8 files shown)


src/sentry_webhook/index.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+  // Use the request message if SentryEvent.message is unpopulated.
+  sentryEvent.message = sentryEvent.message || req.body.message;
+  postSentryEventToSalesforce(sentryEvent, req.body.project)
+      .then(() => {
+        res.status(200).send();
+      })
+      .catch((e) => {
+        console.error(e);
+        // Send an OK response to Sentry - they don't need to know about errors with posting to
```
trevj: You'll hate me for asking but did you test if Sentry will retry a webhook if it receives, e.g, a 500?

src/sentry_webhook/post_sentry_event_to_salesforce.ts
```diff
@@ -0,0 +1,162 @@
...
+  recordType: '0120b0000006e8i',
+};
+
+// Posts a Sentry event to Salesforce using predefined form data. Only posts if the event contains
+// an email address.
+export function postSentryEventToSalesforce(
+    event: sentry.SentryEvent, project: string): Promise<void> {
+  return new Promise((resolve, reject) => {
+    if (!event.user || !event.user.email) {
+      reject(new Error('Not posting event without email'));
```
trevj: Seems like this way we'll get a lot of errors in our logs. How about extracting a function called something like `isSalesforceReport` and calling that first?

src/sentry_webhook/post_sentry_event_to_salesforce.ts
```diff
@@ -0,0 +1,162 @@
...
+  if (!tags) {
+    return null;
+  }
+  const tagsMap = new Map<string, string>();
+  for (const i of Object.keys(tags)) {
+    try {
+      const tagEntry = tags[i];
+      tagsMap.set(tagEntry[0], tagEntry[1]);
+    } catch (e) {
+      console.error('Failed to process tag entry');
```
trevj: Don't swallow the exception.

src/sentry_webhook/post_sentry_event_to_salesforce.ts
```diff
@@ -0,0 +1,162 @@
...
+    formFields: SalesforceFormFields, formValues: SalesforceFormValues, event: sentry.SentryEvent,
+    email: string, isClient: boolean): string {
+  const form = [];
+  form.push(encodeFormData(formFields.orgId, formValues.orgId));
+  form.push(encodeFormData(formFields.recordType, formValues.recordType));
+  form.push(encodeFormData(formFields.email, email));
+  form.push(encodeFormData(formFields.sentryEventId, event.event_id));
+  form.push(encodeFormData(formFields.description, event.message));
+  form.push(encodeFormData(formFields.type, isClient ? 'Outline client' : 'Outline manager'));
+  const tags = getTagsMap(event.tags);
```
trevj: Strongly suggest we:
- make the argument to `getTagsMap` required
- change this to `if (event.tags) { // call getTagsMap }`

src/sentry_webhook/post_sentry_event_to_salesforce.ts
```diff
@@ -0,0 +1,162 @@
...
+    form.push(encodeFormData(formFields.version, tags.get('sentry:release')));
+    if (!isClient) {
+      form.push(encodeFormData(formFields.cloudProvider, tags.get('cloudProvider')));
+    }
+  }
+  return form.join('&');
+}
+
+function encodeFormData(field: string, value?: string) {
+  return `${encodeURIComponent(field)}=${encodeURIComponent(value || '')}`;
```
trevj: FYI `encodeURIComponent` doesn't throw when passed the empty string

src/sentry_webhook/post_sentry_event_to_salesforce.ts
```diff
@@ -0,0 +1,162 @@
...
+        (res) => {
+          if (res.statusCode === 200) {
+            resolve();
+          } else {
+            reject(new Error(`Failed to post form data, response status: ${res.statusCode}`));
+          }
+        });
+    req.on('error', (err) => {
+      reject(new Error(
+          `Failed to submit form: ${err}`,
```
trevj: the trailing comma is messing up the formatting here

src/sentry_webhook/index.ts
alalamav: I did, it does not retry. Yet, I believe we shouldn't be sending them errors unless there is a problem with the request itself.

src/sentry_webhook/post_sentry_event_to_salesforce.ts
alalamav: Done in 4d60e67.

src/sentry_webhook/post_sentry_event_to_salesforce.ts
alalamav: I don't want to fail the whole operation for one malformed tag. I added this as a precaution, though I have not seen malformed tags in any reports.

src/sentry_webhook/post_sentry_event_to_salesforce.ts
alalamav: Good suggestion. Done in 4d60e67.

src/sentry_webhook/post_sentry_event_to_salesforce.ts
alalamav: It encodes `undefined` to `"undefined"` though - rather have an empty string.

src/sentry_webhook/post_sentry_event_to_salesforce.ts
alalamav: Done in 4d60e67.

src/sentry_webhook/index.ts
trevj: If they don't retry, I agree.

src/sentry_webhook/post_sentry_event_to_salesforce.ts
trevj: Mmm, okay. Let's make it a `warn` then, and include some debugging in the message, e.g.:
```ts
console.warn(`failed to process tag entry ${tagEntry[0]}`);
```
(does the environment have `console.warn`? I forget)

PR #361, opened 2018-12-20
Use 127.0.0.1 instead of localhost (refer to #341)

rtsp:
Followup from #341. The `127.0.0.1` fixed the problem but there're 2 other services using `localhost` too. Now I've same problem with those 2 services instead so using `127.0.0.1` instead of `localhost` should fix these problem too.


rtsp:
I signed it!

PR #373, opened 2019-01-07
update electron to 2.0.16, electron-builder 20.34.0

trevj:
https://github.com/electron-userland/electron-builder/releases/tag/v20.34.0 is the last that supports Electron 2.x:
https://github.com/electron-userland/electron-builder/issues/3505#issuecomment-442156509


trevj:
Could be :-)

PR #422, opened 2019-04-17
Remove the extra space.

JonathanDCohen:
I forgot to push the fix last time.


src/server_manager/ui_components/outline-server-settings.html
```diff
@@ -194,7 +194,7 @@ <h3>Share anonymous metrics</h3>
           this.fire('ServerRenameRequested', {newName});
         }
       },
-       _metricsEnabledChanged: function (newMetricsEnabled, oldMetricsEnabled) {
+       _metricsEnabledChanged: function(newMetricsEnabled, oldMetricsEnabled) {
```
alalamav: There also seems to be an extra space before the method name.

JonathanDCohen: Done.

PR #448, opened 2019-06-14
Update for security vulnerabilities

JonathanDCohen:


alalamav:
Nice! I still see js-yaml < 3.13.1, since it's a transitive dependency. I ran into a similar issue in the client; I guess we need to wait until our actual dependencies update it?

PR #456, opened 2019-07-10
Access key quotas

alalamav:
* Implements support for access key (outbound) data transfer quotas over a sliding time window, expressed in hours.
  * Enforces access keys quotas hourly by comparing their usage over the set time window to the set quota. Disables over-quota access keys at the Shadowsocks server and enables them once they are under-quota.
* Exposes APIs to set and remove access key quotas through the management service.
  * `PUT /access-key/{id}/quota`
  * `DELETE /access-key/{id}/quota`
* Adds functionality to query per-access-key hourly data transfer usage through Prometheus metrics.
* Refactors `ServerAccessKeyRepository` to keep access keys in memory and improves asynchronous interface.
* Implements unit tests for every new method.

### DIFF
--- a/src/shadowbox/README.md
+++ b/src/shadowbox/README.md
@@ -122,7 +122,7 @@ curl --insecure -X DELETE $API_URL/access-keys/2
 Set an access key quota
 (e.g. limit outbound data transfer for access key 2 to 1MB over a 24 hour sliding window)
 ```
-curl -v --insecure -X PUT -H "Content-Type: application/json" -d '{"quota": {"quotaBytes": 1000000, "windowHours": 24}}' $API_URL/access-keys/2/quota
+curl -v --insecure -X PUT -H "Content-Type: application/json" -d '{"quota": {"quota": {"bytes": 1000}, "window": {"hours": 1}}}' $API_URL/access-keys/2/quota
 ```
 
 Remove an access key quota
--- a/src/shadowbox/README.md
+++ b/src/shadowbox/README.md
@@ -83,7 +83,7 @@ docker rmi $(docker images -f dangling=true -q)
 
 In order to utilize the Management API, you'll need to know the apiUrl for your Outline server.
 You can obtain this information from the "Settings" tab of the server page in the Outline Manager.
-Alternatively, you can check the 'access.txt' file under the '/opt/outline' directory of an Outline server. An example apiUrl is: https://1.2.3.4:1234/3pQ4jf6qSr5WVeMO0XOo4z. 
+Alternatively, you can check the 'access.txt' file under the '/opt/outline' directory of an Outline server. An example apiUrl is: https://1.2.3.4:1234/3pQ4jf6qSr5WVeMO0XOo4z.
 
 See [Full API Documentation](https://rebilly.github.io/ReDoc/?url=https://raw.githubusercontent.com/Jigsaw-Code/outline-server/master/src/shadowbox/server/api.yml).
 The OpenAPI specification can be found at [api.yml](./api.yml).
@@ -117,6 +117,17 @@ Remove an access key
 (e.g. remove access key 2)
 ```
 curl --insecure -X DELETE $API_URL/access-keys/2
+```
+
+Set an access key quota
+(e.g. limit outbound data transfer for access key 2 to 1MB over a 24 hour sliding window)
+```
+curl -v --insecure -X PUT -H "Content-Type: application/json" -d '{"quota": {"quotaBytes": 1000000, "windowHours": 24}}' $API_URL/access-keys/2/quota
+```
+
+Remove an access key quota
+```
+curl -v --insecure -X DELETE $API_URL/access-keys/2/quota
 ```
 
 ## Testing
--- a/src/shadowbox/README.md
+++ b/src/shadowbox/README.md
@@ -122,7 +122,7 @@ curl --insecure -X DELETE $API_URL/access-keys/2
 Set an access key quota
 (e.g. limit outbound data transfer for access key 2 to 1MB over a 24 hour sliding window)
 ```
-curl -v --insecure -X PUT -H "Content-Type: application/json" -d '{"quota": {"quota": {"bytes": 1000}, "window": {"hours": 1}}}' $API_URL/access-keys/2/quota
+curl -v --insecure -X PUT -H "Content-Type: application/json" -d '{"quota": {"data": {"bytes": 1000}, "window": {"hours": 1}}}' $API_URL/access-keys/2/quota
 ```
 
 Remove an access key quota
--- a/src/shadowbox/model/access_key.ts
+++ b/src/shadowbox/model/access_key.ts
@@ -27,9 +27,9 @@ export interface ProxyParams {
   readonly password: string;
 }
 
-// Parameters needed to enforce an access key data quota, over a sliding window.
+// Parameters needed to enforce an access key data transfer quota, over a sliding window.
 export interface AccessKeyQuota {
-  // The allowed metered traffic measured in bytes.
+  // The allowed metered data transfer measured in bytes.
   readonly quota: {bytes: number};
   // The sliding window size in hours.
   readonly window: {hours: number};
@@ -47,7 +47,7 @@ export interface AccessKey {
   readonly proxyParams: ProxyParams;
   // Admin-controlled, data transfer quota for this access key. Unlimited if unset.
   readonly quota?: AccessKeyQuota;
-  // Whether the access key data usage exceeds the quota.
+  // Whether the access key data transfer exceeds the quota.
   readonly isOverQuota?: boolean;
 }
 
--- a/src/shadowbox/model/access_key.ts
+++ b/src/shadowbox/model/access_key.ts
@@ -27,6 +27,14 @@ export interface ProxyParams {
   password: string;
 }
 
+// Parameters needed to enforce an access key data quota, over a sliding window.
+export interface AccessKeyQuota {
+  // The allowed metered traffic measured in bytes.
+  quotaBytes: number;
+  // The sliding window size in hours.
+  windowHours: number;
+}
+
 // AccessKey is what admins work with. It gives ProxyParams a name and identity.
 export interface AccessKey {
   // The unique identifier for this access key.
@@ -37,6 +45,10 @@ export interface AccessKey {
   metricsId: AccessKeyMetricsId;
   // Parameters to access the proxy
   proxyParams: ProxyParams;
+  // Admin-controlled, data transfer quota for this access key. Unlimited if unset.
+  quota?: AccessKeyQuota;
+  // Whether the access key data usage exceeds the quota.
+  isOverQuota?: boolean;
 }
 
 export interface AccessKeyRepository {
@@ -45,11 +57,14 @@ export interface AccessKeyRepository {
   // Removes the access key given its id.  Returns true if successful.
   removeAccessKey(id: AccessKeyId): boolean;
   // Lists all existing access keys
-  listAccessKeys(): IterableIterator<AccessKey>;
+  listAccessKeys(): AccessKey[];
   // Apply the specified update to the specified access key.
   // Returns true if successful.
   renameAccessKey(id: AccessKeyId, name: string): boolean;
-
   // Gets the metrics id for a given Access Key.
   getMetricsId(id: AccessKeyId): AccessKeyMetricsId|undefined;
+  // Sets the transfer quota for the specified access key. Returns true if successful.
+  setAccessKeyQuota(id: AccessKeyId, quota: AccessKeyQuota): Promise<boolean>;
+  // Clears the transfer quota for the specified access key. Returns true if successful.
+  removeAccessKeyQuota(id: AccessKeyId): Promise<boolean>;
 }
--- a/src/shadowbox/model/access_key.ts
+++ b/src/shadowbox/model/access_key.ts
@@ -55,6 +55,8 @@ export interface AccessKey {
   readonly proxyParams: ProxyParams;
   // Admin-controlled, data transfer quota for this access key. Unlimited if unset.
   readonly quotaUsage?: AccessKeyQuotaUsage;
+  // Returns whether the access key has exceeded its data transfer quota.
+  isOverQuota(): boolean;
 }
 
 export interface AccessKeyRepository {
--- a/src/shadowbox/model/access_key.ts
+++ b/src/shadowbox/model/access_key.ts
@@ -30,9 +30,9 @@ export interface ProxyParams {
 // Parameters needed to enforce an access key data quota, over a sliding window.
 export interface AccessKeyQuota {
   // The allowed metered traffic measured in bytes.
-  readonly quotaBytes: number;
+  readonly quota: {bytes: number};
   // The sliding window size in hours.
-  readonly windowHours: number;
+  readonly window: {hours: number};
 }
 
 // AccessKey is what admins work with. It gives ProxyParams a name and identity.
--- a/src/shadowbox/model/access_key.ts
+++ b/src/shadowbox/model/access_key.ts
@@ -27,12 +27,20 @@ export interface ProxyParams {
   readonly password: string;
 }
 
-// Parameters needed to enforce an access key data transfer quota, over a sliding window.
+// Parameters needed to limit access key data usage over a sliding window.
 export interface AccessKeyQuota {
   // The allowed metered data transfer measured in bytes.
-  readonly quota: {bytes: number};
+  readonly data: {bytes: number};
   // The sliding window size in hours.
   readonly window: {hours: number};
+}
+
+// Parameters needed to enforce an access key data transfer quota.
+export interface AccessKeyQuotaUsage {
+  // Data transfer quota on this access key.
+  readonly quota: AccessKeyQuota;
+  // Data transferred by this access key over the quota window.
+  readonly usage: {bytes: number};
 }
 
 // AccessKey is what admins work with. It gives ProxyParams a name and identity.
@@ -46,9 +54,7 @@ export interface AccessKey {
   // Parameters to access the proxy
   readonly proxyParams: ProxyParams;
   // Admin-controlled, data transfer quota for this access key. Unlimited if unset.
-  readonly quota?: AccessKeyQuota;
-  // Whether the access key data transfer exceeds the quota.
-  readonly isOverQuota?: boolean;
+  readonly quotaUsage?: AccessKeyQuotaUsage;
 }
 
 export interface AccessKeyRepository {
--- a/src/shadowbox/model/access_key.ts
+++ b/src/shadowbox/model/access_key.ts
@@ -18,37 +18,37 @@ export type AccessKeyMetricsId = string;
 // Parameters needed to access a Shadowsocks proxy.
 export interface ProxyParams {
   // Hostname of the proxy
-  hostname: string;
+  readonly hostname: string;
   // Number of the port where the Shadowsocks service is running.
-  portNumber: number;
+  readonly portNumber: number;
   // The Shadowsocks encryption method being used.
-  encryptionMethod: string;
+  readonly encryptionMethod: string;
   // The password for the encryption.
-  password: string;
+  readonly password: string;
 }
 
 // Parameters needed to enforce an access key data quota, over a sliding window.
 export interface AccessKeyQuota {
   // The allowed metered traffic measured in bytes.
-  quotaBytes: number;
+  readonly quotaBytes: number;
   // The sliding window size in hours.
-  windowHours: number;
+  readonly windowHours: number;
 }
 
 // AccessKey is what admins work with. It gives ProxyParams a name and identity.
 export interface AccessKey {
   // The unique identifier for this access key.
-  id: AccessKeyId;
+  readonly id: AccessKeyId;
   // Admin-controlled, editable name for this access key.
-  name: string;
+  readonly name: string;
   // Used in metrics reporting to decouple from the real id. Can change.
-  metricsId: AccessKeyMetricsId;
+  readonly metricsId: AccessKeyMetricsId;
   // Parameters to access the proxy
-  proxyParams: ProxyParams;
+  readonly proxyParams: ProxyParams;
   // Admin-controlled, data transfer quota for this access key. Unlimited if unset.
-  quota?: AccessKeyQuota;
+  readonly quota?: AccessKeyQuota;
   // Whether the access key data usage exceeds the quota.
-  isOverQuota?: boolean;
+  readonly isOverQuota?: boolean;
 }
 
 export interface AccessKeyRepository {
--- a/src/shadowbox/model/shadowsocks_server.ts
+++ b/src/shadowbox/model/shadowsocks_server.ts
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-export interface AccessKey {
+export interface ShadowsocksAccessKey {
   id: string;
   port: number;
   cipher: string;
@@ -21,5 +21,5 @@ export interface AccessKey {
 
 export interface ShadowsocksServer {
   // Updates the server to accept only the given access keys.
-  update(keys: AccessKey[]): Promise<void>;
+  update(keys: ShadowsocksAccessKey[]): Promise<void>;
 }
--- a/src/shadowbox/model/shadowsocks_server.ts
+++ b/src/shadowbox/model/shadowsocks_server.ts
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Parameters required to identify and authenticate connections to a Shadowsocks server.
 export interface ShadowsocksAccessKey {
   id: string;
   port: number;
--- a/src/shadowbox/server/api.yml
+++ b/src/shadowbox/server/api.yml
@@ -8,9 +8,9 @@ tags:
     description: Server-level functions
   - name: Access Key
     description: Access key functions
-servers: 
+servers:
   - url: https://myserver/SecretPath
-    description: Example URL. Change to your own server.  
+    description: Example URL. Change to your own server.
 paths:
   /server:
     get:
@@ -35,7 +35,7 @@ paths:
         - Server
       requestBody:
         required: true
-        content: 
+        content:
           application/json:
             schema:
               type: object
@@ -137,7 +137,7 @@ paths:
             type: string
       requestBody:
         required: true
-        content: 
+        content:
           application/json:
             schema:
               type: object
@@ -147,6 +147,50 @@ paths:
       responses:
         '204':
           description: Access key renamed successfully
+        '404':
+          description: Access key inexistent
+  /access-keys/{id}/quota:
+    put:
+      description: Sets an access key data transfer quota
+      tags:
+        - Access Key
+        - Quota
+      parameters:
+        - name: id
+          in: path
+          required: true
+          description: The id of the access key to set a quota
+          schema:
+            type: string
+      requestBody:
+        required: true
+        content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/AccessKeyQuota"
+              examples:
+                '0':
+                  value: "{quotaBytes: 1000000, windowHours: 24}"
+      responses:
+        '204':
+          description: Access key quota set successfully
+        '404':
+          description: Access key inexistent
+    delete:
+      description: Removes an acess key data transfer quota
+      tags:
+        - Access Key
+        - Quota
+      parameters:
+        - name: id
+          in: path
+          required: true
+          description: The id of the access key to delete a quota
+          schema:
+            type: string
+      responses:
+        '204':
+          description: Access key quota set successfully
         '404':
           description: Access key inexistent
   /metrics/transfer:
@@ -193,7 +237,7 @@ paths:
         - Server
       requestBody:
         required: true
-        content: 
+        content:
           application/json:
             schema:
               type: object
@@ -208,7 +252,7 @@ paths:
           description: Setting successful
         '400':
           description: Invalid request
-          
+
 components:
   schemas:
     Server:
@@ -222,6 +266,12 @@ components:
         createdTimestampMs:
           type: number
         portForNewAccessKeys:
+          type: integer
+    AccessKeyQuota:
+      properties:
+        quotaBytes:
+          type: integer
+        windowHours:
           type: integer
     AccessKey:
       required:
@@ -239,3 +289,5 @@ components:
           type: string
         accessUrl:
           type: string
+        quota:
+          $ref: "#/components/schemas/AccessKeyQuota"
--- a/src/shadowbox/server/api.yml
+++ b/src/shadowbox/server/api.yml
@@ -269,10 +269,16 @@ components:
           type: integer
     AccessKeyQuota:
       properties:
-        quotaBytes:
-          type: integer
-        windowHours:
-          type: integer
+        quota:
+          type: object
+          properties:
+            bytes:
+              type: integer
+        window:
+          type: object
+          properties:
+            hours:
+              type: integer
     AccessKey:
       required:
         - id
--- a/src/shadowbox/server/api.yml
+++ b/src/shadowbox/server/api.yml
@@ -269,7 +269,7 @@ components:
           type: integer
     AccessKeyQuota:
       properties:
-        quota:
+        data:
           type: object
           properties:
             bytes:
--- a/src/shadowbox/server/api.yml
+++ b/src/shadowbox/server/api.yml
@@ -177,7 +177,7 @@ paths:
         '404':
           description: Access key inexistent
     delete:
-      description: Removes an acess key data transfer quota
+      description: Removes an access key data transfer quota, lifting data transfer restrictions on the key.
       tags:
         - Access Key
         - Quota
@@ -190,7 +190,7 @@ paths:
             type: string
       responses:
         '204':
-          description: Access key quota set successfully
+          description: Access key quota deleted successfully.
         '404':
           description: Access key inexistent
   /metrics/transfer:
--- a/src/shadowbox/server/manager_metrics.ts
+++ b/src/shadowbox/server/manager_metrics.ts
@@ -17,7 +17,7 @@ import {DataUsageByUser} from '../model/metrics';
 
 export interface ManagerMetrics {
   get30DayByteTransfer(): Promise<DataUsageByUser>;
-  getOutboundByteTransfer(accessKeyId: string, windowHours: number): Promise<DataUsageByUser>;
+  getOutboundByteTransfer(accessKeyId: string, windowHours: number): Promise<number>;
 }
 
 // Reads manager metrics from a Prometheus instance.
@@ -42,17 +42,14 @@ export class PrometheusManagerMetrics implements ManagerMetrics {
     return {bytesTransferredByUserId: usage};
   }
 
-  async getOutboundByteTransfer(accessKeyId: string, windowHours: number):
-      Promise<DataUsageByUser> {
+  async getOutboundByteTransfer(accessKeyId: string, windowHours: number): Promise<number> {
     let bytesTransferred = 0;
     const result = await this.prometheusClient.query(
-        `sum(increase(shadowsocks_data_bytes{dir=~"c<p|p>t|>p<",access_key="${accessKeyId}"}[${
+        `sum(increase(shadowsocks_data_bytes{dir=~"c<p|p>t",access_key="${accessKeyId}"}[${
             windowHours}h])) by (access_key)`);
     if (result && result.result[0] && result.result[0].metric['access_key'] === accessKeyId) {
       bytesTransferred = Math.round(parseFloat(result.result[0].value[1]));
     }
-    const bytesTransferredByUserId = {};
-    bytesTransferredByUserId[accessKeyId] = bytesTransferred;
-    return {bytesTransferredByUserId};
+    return bytesTransferred;
   }
 }
--- a/src/shadowbox/server/manager_metrics.ts
+++ b/src/shadowbox/server/manager_metrics.ts
@@ -15,7 +15,10 @@
 import {PrometheusClient} from '../infrastructure/prometheus_scraper';
 import {DataUsageByUser} from '../model/metrics';
 
-export interface ManagerMetrics { get30DayByteTransfer(): Promise<DataUsageByUser>; }
+export interface ManagerMetrics {
+  get30DayByteTransfer(): Promise<DataUsageByUser>;
+  getOutboundByteTransfer(accessKeyId: string, windowHours: number): Promise<DataUsageByUser>;
+}
 
 // Reads manager metrics from a Prometheus instance.
 export class PrometheusManagerMetrics implements ManagerMetrics {
@@ -37,5 +40,19 @@ export class PrometheusManagerMetrics implements ManagerMetrics {
       usage[entry.metric['access_key'] || ''] = bytes;
     }
     return {bytesTransferredByUserId: usage};
+  }
+
+  async getOutboundByteTransfer(accessKeyId: string, windowHours: number):
+      Promise<DataUsageByUser> {
+    let bytesTransferred = 0;
+    const result = await this.prometheusClient.query(
+        `sum(increase(shadowsocks_data_bytes{dir=~"c<p|p>t|>p<",access_key="${accessKeyId}"}[${
+            windowHours}h])) by (access_key)`);
+    if (result && result.result[0] && result.result[0].metric['access_key'] === accessKeyId) {
+      bytesTransferred = Math.round(parseFloat(result.result[0].value[1]));
+    }
+    const bytesTransferredByUserId = {};
+    bytesTransferredByUserId[accessKeyId] = bytesTransferred;
+    return {bytesTransferredByUserId};
   }
 }
--- a/src/shadowbox/server/manager_metrics.ts
+++ b/src/shadowbox/server/manager_metrics.ts
@@ -17,7 +17,6 @@ import {DataUsageByUser} from '../model/metrics';
 
 export interface ManagerMetrics {
   get30DayByteTransfer(): Promise<DataUsageByUser>;
-  getOutboundByteTransfer(accessKeyId: string, windowHours: number): Promise<number>;
 }
 
 // Reads manager metrics from a Prometheus instance.
@@ -40,16 +39,5 @@ export class PrometheusManagerMetrics implements ManagerMetrics {
       usage[entry.metric['access_key'] || ''] = bytes;
     }
     return {bytesTransferredByUserId: usage};
-  }
-
-  async getOutboundByteTransfer(accessKeyId: string, windowHours: number): Promise<number> {
-    let bytesTransferred = 0;
-    const result = await this.prometheusClient.query(
-        `sum(increase(shadowsocks_data_bytes{dir=~"c<p|p>t",access_key="${accessKeyId}"}[${
-            windowHours}h])) by (access_key)`);
-    if (result && result.result[0] && result.result[0].metric['access_key'] === accessKeyId) {
-      bytesTransferred = Math.round(parseFloat(result.result[0].value[1]));
-    }
-    return bytesTransferred;
   }
 }
--- a/src/shadowbox/server/manager_service.ts
+++ b/src/shadowbox/server/manager_service.ts
@@ -17,7 +17,7 @@ import {makeConfig, SIP002_URI} from 'ShadowsocksConfig/shadowsocks_config';
 
 import {JsonConfig} from '../infrastructure/json_config';
 import * as logging from '../infrastructure/logging';
-import {AccessKey, AccessKeyRepository} from '../model/access_key';
+import {AccessKey, AccessKeyQuota, AccessKeyRepository} from '../model/access_key';
 
 import {ManagerMetrics} from './manager_metrics';
 import {ServerConfigJson} from './server_config';
@@ -40,7 +40,9 @@ function accessKeyToJson(accessKey: AccessKey) {
       method: accessKey.proxyParams.encryptionMethod,
       password: accessKey.proxyParams.password,
       outline: 1,
-    }))
+    })),
+    quota: accessKey.quota,
+    isOverQuota: accessKey.isOverQuota || false
   };
 }
 
@@ -50,6 +52,7 @@ interface RequestParams {
   id?: string;
   name?: string;
   metricsEnabled?: boolean;
+  quota?: AccessKeyQuota;
 }
 interface RequestType {
   params: RequestParams;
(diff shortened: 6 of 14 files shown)


src/shadowbox/server/manager_service.ts
```diff
@@ -168,6 +173,46 @@ export class ShadowsocksManagerService {
     }
   }
 
+  public async setAccessKeyQuota(req: RequestType, res: ResponseType, next: restify.Next) {
+    try {
+      logging.debug(`setAccessKeyQuota request ${JSON.stringify(req.params)}`);
+      const accessKeyId = req.params.id;
```
JonathanDCohen: Is it possible that id is undefined?  It doesn't look like we handle that case

src/shadowbox/server/server_access_key.ts
```diff
@@ -56,28 +61,56 @@ function makeAccessKey(hostname: string, accessKeyJson: AccessKeyConfig): Access
...
   };
 }
 
 // AccessKeyRepository that keeps its state in a config file and uses ShadowsocksServer
 // to start and stop per-access-key Shadowsocks instances.
 export class ServerAccessKeyRepository implements AccessKeyRepository {
-  // This is the max id + 1 among all access keys. Used to generate unique ids for new access keys.
   private NEW_USER_ENCRYPTION_METHOD = 'chacha20-ietf-poly1305';
+  private QUOTA_ENFORCEMENT_INTERVAL_MS = 3600000;  // 1h
+  private accessKeys: AccessKey[];
```
JonathanDCohen: Is there a particular reason why we don't use a Set\<AccessKey\>?

src/shadowbox/server/server_access_key.ts
```diff
@@ -90,73 +123,162 @@ export class ServerAccessKeyRepository implements AccessKeyRepository {
...
 
-  private getAccessKey(id: AccessKeyId): AccessKeyConfig {
-    for (const accessKeyJson of this.keyConfig.data().accessKeys) {
-      if (accessKeyJson.id === id) {
-        return accessKeyJson;
+  private loadAccessKeys(): AccessKey[] {
+    return this.keyConfig.data().accessKeys.map(e => makeAccessKey(this.proxyHostname, e));
+  }
+
+  private saveAccessKeys() {
```
JonathanDCohen: maybe (de)?serializeAccessKeys() is more descriptive?

src/shadowbox/server/server_access_key.ts
```diff
@@ -90,73 +123,162 @@ export class ServerAccessKeyRepository implements AccessKeyRepository {
...
+    }
+    const quotaStausChanged = accessKey.isOverQuota;
+    accessKey.quota = undefined;
+    accessKey.isOverQuota = false;
+    try {
+      this.saveAccessKeys();
+      if (quotaStausChanged) {
+        await this.updateServer();
+      }
     } catch (error) {
```
JonathanDCohen: I'd like to know a little more about the possible errors here.  Specifically, is it possible and how can our serialized list of AccessKeys get out of sync with the expectations of the manager?  Is it basically just on a file write error?  I guess also more generally, is there a way we expect callers to handle this error other than retrying?  It seems to me like if we can't write a JSON file then there are som ebigger problems happening somewhere

src/shadowbox/server/server_access_key.ts
```diff
@@ -90,73 +123,162 @@ export class ServerAccessKeyRepository implements AccessKeyRepository {
...
+    const quotaStatusChanged = isOverQuota !== accessKey.isOverQuota;
+    accessKey.isOverQuota = isOverQuota;
+    return quotaStatusChanged;
   }
 
   private updateServer(): Promise<void> {
-    return this.shadowsocksServer.update(this.keyConfig.data().accessKeys.map((e) => {
-      return {id: e.id, port: e.port, cipher: e.encryptionMethod, secret: e.password};
-    }));
+    const serverAccessKeys = this.accessKeys.filter(e => !e.isOverQuota).map(e => {
```
JonathanDCohen: It might be nice to take this opportunity to use "key" or something more descriptive than "e"

src/shadowbox/server/server_access_key.ts
```diff
@@ -90,73 +123,162 @@ export class ServerAccessKeyRepository implements AccessKeyRepository {
...
+  // the quota status changed.
+  private async updateAccessKeyQuotaStaus(accessKey: AccessKey): Promise<boolean> {
+    if (!accessKey.quota) {
+      return false;  // Don't query the usage of access keys without quota.
+    }
+    const usage =
+        await this.metrics.getOutboundByteTransfer(accessKey.id, accessKey.quota.windowHours);
+    const usageBytes = usage.bytesTransferredByUserId[accessKey.id];
+    const isOverQuota = usageBytes > accessKey.quota.quotaBytes;
+    logging.debug(`Enforcing quota for access key ${accessKey.id}. Quota: ${
```
JonathanDCohen: Nit: do we care that logging on a quota which the key hasn't met might pollute our logs?

src/shadowbox/server/server_access_key.ts
```diff
@@ -90,73 +123,162 @@ export class ServerAccessKeyRepository implements AccessKeyRepository {
...
-    const accessKeyJson = this.getAccessKey(id);
-    return accessKeyJson ? accessKeyJson.metricsId : undefined;
+    const accessKey = this.getAccessKey(id);
+    return accessKey ? accessKey.metricsId : undefined;
+  }
+
+  // Compares access key usage with collected metrics, marking them as under or over quota.
+  async enforceAccessKeyQuotas() {
+    let quotaStatusChanged = false;
+    for (const accessKey of this.accessKeys) {
```
JonathanDCohen: I wonder if this might be an accidental pessimization to query prometheus a bunch of times for each key as opposed to just calling updateServer() and being okay if nothing changed.

src/shadowbox/server/server_access_key.ts
```diff
@@ -90,73 +123,162 @@ export class ServerAccessKeyRepository implements AccessKeyRepository {
...
+      return false;
+    }
+    const accessKey = this.getAccessKey(id);
+    if (!accessKey) {
+      return false;
+    }
+    accessKey.quota = quota;
+    try {
+      this.saveAccessKeys();
+      const quotaStausChanged = await this.updateAccessKeyQuotaStaus(accessKey);
```
JonathanDCohen: s/Staus/Status

src/shadowbox/server/server_access_key.ts
```diff
@@ -90,73 +123,162 @@ export class ServerAccessKeyRepository implements AccessKeyRepository {
...
-    return accessKeyJson ? accessKeyJson.metricsId : undefined;
+    const accessKey = this.getAccessKey(id);
+    return accessKey ? accessKey.metricsId : undefined;
+  }
+
+  // Compares access key usage with collected metrics, marking them as under or over quota.
+  async enforceAccessKeyQuotas() {
+    let quotaStatusChanged = false;
+    for (const accessKey of this.accessKeys) {
+      quotaStatusChanged = quotaStatusChanged || await this.updateAccessKeyQuotaStaus(accessKey);
```
JonathanDCohen: s/Staus/Status

src/shadowbox/server/server_access_key.ts
```diff
@@ -90,73 +123,162 @@ export class ServerAccessKeyRepository implements AccessKeyRepository {
...
+    }
+    if (quotaStatusChanged) {
+      this.updateServer();
+    }
+  }
+
+  // Updates `accessKey` quota status by comparing its usage with collected metrics. Returns whether
+  // the quota status changed.
+  private async updateAccessKeyQuotaStaus(accessKey: AccessKey): Promise<boolean> {
```
JonathanDCohen: s/Staus/Status

src/shadowbox/server/server_access_key.ts
```diff
@@ -90,73 +123,162 @@ export class ServerAccessKeyRepository implements AccessKeyRepository {
...
+    }
+    return true;
+  }
+
+  async removeAccessKeyQuota(id: AccessKeyId): Promise<boolean> {
+    const accessKey = this.getAccessKey(id);
+    if (!accessKey) {
+      return false;
+    }
+    const quotaStausChanged = accessKey.isOverQuota;
```
JonathanDCohen: This variable name seems off to me.

src/shadowbox/server/api.yml
```diff
@@ -149,6 +149,50 @@ paths:
...
+      parameters:
+        - name: id
+          in: path
+          required: true
+          description: The id of the access key to delete a quota
+          schema:
+            type: string
+      responses:
+        '204':
+          description: Access key quota set successfully
```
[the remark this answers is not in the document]
alalamav: Good catch, done in 2105d3f.

src/shadowbox/model/shadowsocks_server.ts
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-export interface AccessKey {
+export interface ShadowsocksAccessKey {
```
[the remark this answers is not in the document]
alalamav: Sounds right. The way I see it, they both provide access to a Shadowsocks server from a different perspective. `AccessKey` is what the server manager works with, hence the additional metadata such as the name, quota, etc. `ShadowsocksAccessKey` is what the server needs to authenticate and identify connections.

I have added a comment in 2105d3f to clarify the latter.

src/shadowbox/server/api.yml
```diff
@@ -149,6 +149,50 @@ paths:
...
+              examples:
+                '0':
+                  value: "{quotaBytes: 1000000, windowHours: 24}"
+      responses:
+        '204':
+          description: Access key quota set successfully
+        '404':
+          description: Access key inexistent
+    delete:
+      description: Removes an acess key data transfer quota
```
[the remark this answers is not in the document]
alalamav: Done, expanded the comment in 2105d3f.

src/shadowbox/server/server_access_key.ts
```diff
@@ -56,28 +61,56 @@ function makeAccessKey(hostname: string, accessKeyJson: AccessKeyConfig): Access
...
     }
     if (this.keyConfig.data().nextId === undefined) {
       this.keyConfig.data().nextId = 0;
     }
-    this.updateServer();
+    this.accessKeys = this.loadAccessKeys();
+  }
+
+  // Exposes the access key configuration to the server. Periodically enforces access key quotas.
+  async start(clock: Clock): Promise<void> {
```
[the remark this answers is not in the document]
alalamav: I agree with you and actually considered naming it `startQuotaEnforcement`; however, the method does more than enforcing the quotas, it exposes the keys to the server. 
We could call `updateServer` in the constructor, reverting to the semi-asynchronous behavior, rename this method, and have it only enforce quotas. WDYT?

src/shadowbox/server/manager_service.ts
alalamav: It is, in which case the repo will return false and we will serve a 404.

src/shadowbox/server/server_access_key.ts
alalamav: One I can think of is indexed access being useful in tests.

src/shadowbox/server/server_access_key.ts
alalamav: I wanted the name to imply a write/read operation, thus I chose 'save'/'load'. The (de)serialization seems more of an implementation detail.

src/shadowbox/server/server_access_key.ts
alalamav: Done in 2105d3f.

src/shadowbox/server/server_access_key.ts
alalamav: Debug logging shouldn't appear in production servers.

src/shadowbox/server/server_access_key.ts
alalamav: We only query Prometheus once for each key that has a quota. I don't think the extra measure to update the server only when necessary overly complicates the code, but I would be okay with calling `updateServer` unconditionally if you disagree.

Please let me know if I misunderstood your point.

src/shadowbox/server/server_access_key.ts
alalamav: Done in 2105d3f.

src/shadowbox/server/server_access_key.ts
alalamav: This one is actually okay :)

src/shadowbox/server/server_access_key.ts
alalamav: Done in 2105d3f - hard word to spell 😅

src/shadowbox/server/server_access_key.ts
alalamav: Renamed to `wasOverQuota` in 2105d3f, which is more specific.

src/shadowbox/server/server_access_key.ts
alalamav: `update` can fail due to a write error or a problem starting the shadowsocks process. I've been thinking about this too. IMO, we need typed errors in the access key repository. This way, the caller can take the appropriate actions (i.e. retry, panic).

I decided not to takle this refactor on this PR, since it is already large.

src/shadowbox/server/server_access_key.ts
JonathanDCohen: Also it looks like Sets of object types are pretty janky in Typescript since it uses identity to compare for equality.  So never mind on that one :)

src/shadowbox/model/access_key.ts
```diff
@@ -27,6 +27,14 @@ export interface ProxyParams {
   password: string;
 }
 
+// Parameters needed to enforce an access key data quota, over a sliding window.
+export interface AccessKeyQuota {
+  // The allowed metered traffic measured in bytes.
+  quotaBytes: number;
+  // The sliding window size in hours.
+  windowHours: number;
```
fortuna: Fixing hours will limit us. I believe most admins will think in terms of days. We don't want them to enter days in hours or display days in hours. We should be able to let the admin choose the unit, and save that.

The same applies to the data field. We should be able to specify the unit. Entering bytes is error prone.

What's the proposed solution?

I guess one possibility is to allow the admin to pick the unit on input, but save the values in this normalized format. Then picking a "best unit" when displaying.

src/shadowbox/model/access_key.ts
```diff
@@ -37,6 +45,10 @@ export interface AccessKey {
   metricsId: AccessKeyMetricsId;
   // Parameters to access the proxy
   proxyParams: ProxyParams;
+  // Admin-controlled, data transfer quota for this access key. Unlimited if unset.
+  quota?: AccessKeyQuota;
```
fortuna: How do we represent "default" quota?

src/shadowbox/server/manager_service.ts
```diff
@@ -168,6 +173,46 @@ export class ShadowsocksManagerService {
...
+      if (!quota || !quota.quotaBytes || !quota.windowHours) {
+        return next(new restify.InvalidArgumentError(
+            'Must provide a quota value with "quotaBytes" and "windowHours"'));
+      }
+      if (quota.quotaBytes < 0 || quota.windowHours < 0) {
+        return next(new restify.InvalidArgumentError('Must provide positive quota values'));
+      }
+      const success = await this.accessKeys.setAccessKeyQuota(accessKeyId, quota);
+      if (!success) {
+        return next(new restify.NotFoundError(`No access key found with id ${accessKeyId}`));
```
fortuna: The method could potentially fail on invalid input, no?

src/shadowbox/server/server_access_key.ts
```diff
@@ -90,73 +123,162 @@ export class ServerAccessKeyRepository implements AccessKeyRepository {
...
       }
     }
     return false;
   }
 
-  listAccessKeys(): IterableIterator<AccessKey> {
-    return this.keyConfig.data().accessKeys.map(
-        accessKeyJson => makeAccessKey(this.proxyHostname, accessKeyJson))[Symbol.iterator]();
+  listAccessKeys(): AccessKey[] {
+    // Deep copy the access key array to prevent returning references to keys or their properties.
```
fortuna: Why is it bad to keep references to the keys?

src/shadowbox/server/manager_metrics.ts
```diff
@@ -38,4 +41,18 @@ export class PrometheusManagerMetrics implements ManagerMetrics {
...
+    let bytesTransferred = 0;
+    const result = await this.prometheusClient.query(
+        `sum(increase(shadowsocks_data_bytes{dir=~"c<p|p>t|>p<",access_key="${accessKeyId}"}[${
+            windowHours}h])) by (access_key)`);
+    if (result && result.result[0] && result.result[0].metric['access_key'] === accessKeyId) {
+      bytesTransferred = Math.round(parseFloat(result.result[0].value[1]));
+    }
+    const bytesTransferredByUserId = {};
+    bytesTransferredByUserId[accessKeyId] = bytesTransferred;
+    return {bytesTransferredByUserId};
```
JonathanDCohen: Why do we wrap it in another object?  Could we

`return {accessKeyId: bytesTransferred}` or even just `return bytesTransferred` directly?

src/shadowbox/server/mocks/mocks.ts
```diff
@@ -78,3 +35,35 @@ export class InMemoryFile implements TextFile {
...
+    this.accessKeys = keys;
+    return Promise.resolve();
+  }
+
+  getAccessKeys() {
+    return this.accessKeys;
+  }
+}
+
+export class ManagerMetricsStub implements ManagerMetrics {
```
JonathanDCohen: I believe this is also a fake by Google terminology, since for that a stub "only returns what you tell it to return", whereas this is a shallow implementation of the interface.

src/shadowbox/server/server_access_key.ts
JonathanDCohen: Cool.

src/shadowbox/model/shadowsocks_server.ts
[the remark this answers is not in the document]
JonathanDCohen: Maybe let's call them ShadowsocksAccessKey and OutlineAccessKey then?

src/shadowbox/README.md
```diff
@@ -119,6 +119,17 @@ Remove an access key
...
+(e.g. limit outbound data transfer for access key 2 to 1MB over a 24 hour sliding window)
+```
+curl -v --insecure -X PUT -H "Content-Type: application/json" -d '{"quota": {"quotaBytes": 1000000, "windowHours": 24}}' $API_URL/access-keys/2/quota
+```
+
+Remove an access key quota
+```
+curl -v --insecure -X DELETE $API_URL/access-keys/2/quota
+```
+
```
JonathanDCohen: @fortuna seems like here we would want a PUT access-keys/default-quota here.  Do we know that admins want per-key quotas or is it more likely that the ease of the same quota for everyone is what is called for? 

What I'm thinking is this: right now I believe that access key ids are randomly generated.  If that's the case then doing something reasonable like "change everyone's quota" becomes a bit of a pain because you will have to enumerate and iterate over your access keys.  It also requires a linear number of requests as opposed to a single request.

src/shadowbox/server/manager_metrics.ts
```diff
@@ -38,4 +41,18 @@ export class PrometheusManagerMetrics implements ManagerMetrics {
     }
     return {bytesTransferredByUserId: usage};
   }
+
+  async getOutboundByteTransfer(accessKeyId: string, windowHours: number):
+      Promise<DataUsageByUser> {
+    let bytesTransferred = 0;
+    const result = await this.prometheusClient.query(
+        `sum(increase(shadowsocks_data_bytes{dir=~"c<p|p>t|>p<",access_key="${accessKeyId}"}[${
```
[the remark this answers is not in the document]
alalamav: Done in 2808f76.

src/shadowbox/server/manager_metrics.ts
```diff
@@ -38,4 +41,18 @@ export class PrometheusManagerMetrics implements ManagerMetrics {
     }
     return {bytesTransferredByUserId: usage};
   }
+
+  async getOutboundByteTransfer(accessKeyId: string, windowHours: number):
+      Promise<DataUsageByUser> {
```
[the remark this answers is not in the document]
alalamav: I was going for consistency in the interface, but I now realize it makes more sense to return a number. Fixed in 2808f76.

src/shadowbox/server/manager_metrics.ts
```diff
@@ -38,4 +41,18 @@ export class PrometheusManagerMetrics implements ManagerMetrics {
     }
     return {bytesTransferredByUserId: usage};
   }
+
+  async getOutboundByteTransfer(accessKeyId: string, windowHours: number):
```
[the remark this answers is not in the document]
alalamav: I talked to @cjhenck about this and we agreed to measure outbound traffic, since that is what the admin pays and generally counts towards the server's data usage.

> This is one thing that we may want Annabell to answer.

+1

src/shadowbox/server/server_access_key.ts
```diff
@@ -56,28 +61,56 @@ function makeAccessKey(hostname: string, accessKeyJson: AccessKeyConfig): Access
...
     if (this.keyConfig.data().nextId === undefined) {
       this.keyConfig.data().nextId = 0;
     }
-    this.updateServer();
+    this.accessKeys = this.loadAccessKeys();
+  }
+
+  // Exposes the access key configuration to the server. Periodically enforces access key quotas.
+  async start(clock: Clock): Promise<void> {
+    await this.updateServer();
```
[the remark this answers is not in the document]
alalamav: Yes, moved the `updateServer` call to the end of the method in 2808f76.

src/shadowbox/server/server_access_key.ts
```diff
@@ -56,28 +61,56 @@ function makeAccessKey(hostname: string, accessKeyJson: AccessKeyConfig): Access
...
+    quota: accessKey.quota
   };
 }
 
 // AccessKeyRepository that keeps its state in a config file and uses ShadowsocksServer
 // to start and stop per-access-key Shadowsocks instances.
 export class ServerAccessKeyRepository implements AccessKeyRepository {
-  // This is the max id + 1 among all access keys. Used to generate unique ids for new access keys.
   private NEW_USER_ENCRYPTION_METHOD = 'chacha20-ietf-poly1305';
+  private QUOTA_ENFORCEMENT_INTERVAL_MS = 3600000;  // 1h
```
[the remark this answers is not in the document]
alalamav: Done in 2808f76.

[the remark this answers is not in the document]
alalamav: Done in 2808f76.

src/shadowbox/model/access_key.ts
alalamav: > I guess one possibility is to allow the admin to pick the unit on input, but save the values in this normalized format. Then picking a "best unit" when displaying.

That's what I was thinking. The server manager can display any unit to the user and convert them when sending the request. It can automatically pick the best unit, or persist the user choice locally. 

These units are as fine-grained as we want to support (lowest common denominator). The alternative would be to add unit fields for both parameters, but I think it would add complexity the quota interface and API.

src/shadowbox/model/access_key.ts
alalamav: The default quota should be a server property, persisted in the server config. We agreed that default quotas would be applied to new access keys. I'm thinking we could add an `isDefault` property to `AccessKeyQuota`. We would update the quotas where this property is true when setting the default quota.

Stepping back on previous assumptions, the default quota behavior may benefit from UX research. I plan to implement default quotas in a follow-up PR.

src/shadowbox/server/manager_service.ts
alalamav: It could, hence I added (duplicate) input checks above to serve a 409. Repository methods can also fail due to I/O, but we currenlty have no way of differentiating. I believe we need to refactor the server repository to throw typed errors that allow the caller to take the appropriate action based on different failure conditions. 

I added a TODO to support typed errors in the access key repository in 2808f76.

src/shadowbox/server/server_access_key.ts
alalamav: If we return references to the keys, a caller could bypass the repository's methods to modify them.

src/shadowbox/README.md
alalamav: IMO, the API should be at the server level: /default-quota. See  https://github.com/Jigsaw-Code/outline-server/pull/456#discussion_r303624942

src/shadowbox/server/server_access_key.ts
```diff
@@ -90,73 +123,162 @@ export class ServerAccessKeyRepository implements AccessKeyRepository {
...
     }
     return false;
   }
 
-  listAccessKeys(): IterableIterator<AccessKey> {
-    return this.keyConfig.data().accessKeys.map(
-        accessKeyJson => makeAccessKey(this.proxyHostname, accessKeyJson))[Symbol.iterator]();
+  listAccessKeys(): AccessKey[] {
+    // Deep copy the access key array to prevent returning references to keys or their properties.
+    return this.accessKeys.map(key => (key instanceof Object) ? {...key} : key);
```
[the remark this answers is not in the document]
alalamav: When the type is a primitive.

src/shadowbox/server/server_access_key.spec.ts
```diff
@@ -76,77 +84,279 @@ describe('ServerAccessKeyRepository', () => {
...
+        new InMemoryConfig<AccessKeyConfigJson>({accessKeys: [], nextId: 0}), server, metrics);
+    const accessKey1 = await repo.createNewAccessKey();
+    const accessKey2 = await repo.createNewAccessKey();
+
+    await repo.setAccessKeyQuota(accessKey1.id, {quotaBytes: 200, windowHours: 1});
+    let accessKeys = await repo.listAccessKeys();
+    expect(accessKeys[0].isOverQuota).toBeTruthy();
+    expect(accessKeys[1].isOverQuota).toBeFalsy();
+    let serverAccessKeys = server.getAccessKeys();
+    expect(serverAccessKeys.length).toEqual(1);
```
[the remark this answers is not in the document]
alalamav: `repo.listAccessKeys` returns the access keys from the repo's perspective (includes over-quota keys). `server.getAccessKeys` returns the access keys from the server's perspective after a call to `server.update(accessKeys)` (excludes over-quota keys).

src/shadowbox/server/server_access_key.ts
```diff
@@ -56,28 +61,56 @@ function makeAccessKey(hostname: string, accessKeyJson: AccessKeyConfig): Access
...
   private NEW_USER_ENCRYPTION_METHOD = 'chacha20-ietf-poly1305';
+  private QUOTA_ENFORCEMENT_INTERVAL_MS = 3600000;  // 1h
+  private accessKeys: AccessKey[];
   private portForNewAccessKeys: number|undefined;
 
   constructor(
       private portProvider: PortProvider, private proxyHostname: string,
       private keyConfig: JsonConfig<AccessKeyConfigJson>,
-      private shadowsocksServer: ShadowsocksServer) {
+      private shadowsocksServer: ShadowsocksServer, private metrics: ManagerMetrics) {
```
fortuna: You can pass the Prometheus client here instead, or some abstraction for the quota calculation.

src/shadowbox/server/manager_metrics.ts
```diff
@@ -15,7 +15,10 @@
 import {PrometheusClient} from '../infrastructure/prometheus_scraper';
 import {DataUsageByUser} from '../model/metrics';
 
-export interface ManagerMetrics { get30DayByteTransfer(): Promise<DataUsageByUser>; }
+export interface ManagerMetrics {
+  get30DayByteTransfer(): Promise<DataUsageByUser>;
+  getOutboundByteTransfer(accessKeyId: string, windowHours: number): Promise<DataUsageByUser>;
```
fortuna: This seems to violate the single responsibility rule. I don't see a good reason to conflate the metrics we show to the manager with the quota enforcement system.

Let's separate those things.

Maybe you can place the logic in the `ServerAccessKeyRepository`. Or in an abstraction that lives there. You could inject the Prometheus client in the constructor, rather than the ManagerMetrics.

BTW, the main reason ManagerMetrics exists is because of the transition from the old to the new metrics when we migrated to Go.

src/shadowbox/server/server_access_key.spec.ts
```diff
@@ -76,77 +84,279 @@ describe('ServerAccessKeyRepository', () => {
...
+      // List keys again and expect to NOT see the NEW_NAME.
+      const accessKeys = repo.listAccessKeys();
       expect(accessKeys[0].name).not.toEqual(NEW_NAME);
       done();
     });
   });
 
-  it('Repos created with an existing file restore access keys', (done) => {
+  it('Creates access keys with no quota', async (done) => {
+    const repo = createRepo();
```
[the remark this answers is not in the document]
alalamav: Done in a9bd8a5.

src/shadowbox/server/server_access_key.spec.ts
```diff
@@ -76,77 +84,279 @@ describe('ServerAccessKeyRepository', () => {
...
+    const quota = {quotaBytes: 1000, windowHours: 24};
+    expect(await repo.setAccessKeyQuota('doesnotexist', quota)).toBeFalsy();
+    done();
+  });
+
+  it('setAccessKeyQuota fails with disallowed quota values', async (done) => {
+    const repo = createRepo();
+    const accessKey = await repo.createNewAccessKey();
+    // Negative values
+    let quota = {quotaBytes: -1000, windowHours: 24};
```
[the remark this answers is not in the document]
alalamav: Done in a9bd8a5.

src/shadowbox/server/server_access_key.spec.ts
```diff
@@ -76,77 +84,279 @@ describe('ServerAccessKeyRepository', () => {
...
+
+    // Create a 2nd repo from the same config file. This simulates what
+    // might happen after the shadowbox server is restarted.
+    const repo2 = new ServerAccessKeyRepository(
+        new PortProvider(), 'hostname', config, new MockShadowsocksServer(),
+        new ManagerMetricsStub({}));
+    // Check that repo1 and repo2 have the same access keys
+    const repo1Keys = repo1.listAccessKeys();
+    const repo2Keys = repo2.listAccessKeys();
+    expect(repo1Keys.length).toEqual(2);
```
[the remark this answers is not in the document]
alalamav: Good suggestion. Done in a9bd8a5.

src/shadowbox/server/mocks/mocks.ts
```diff
@@ -78,3 +35,35 @@ export class InMemoryFile implements TextFile {
     this.exists = true;
   }
 }
+
+export class MockShadowsocksServer implements ShadowsocksServer {
```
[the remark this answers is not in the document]
alalamav: Renamed in a9bd8a5.

src/shadowbox/server/mocks/mocks.ts
alalamav: Renamed in a9bd8a5.

src/shadowbox/server/server_access_key.spec.ts
```diff
@@ -76,77 +84,279 @@ describe('ServerAccessKeyRepository', () => {
...
+    expect(serverAccessKeys.length).toEqual(2);
+    expect(serverAccessKeys[0].id).toEqual(accessKey1.id);
+    expect(serverAccessKeys[1].id).toEqual(accessKey2.id);
+    done();
+  });
 
-function countAccessKeys(repo: AccessKeyRepository) {
-  return iterToArray(repo.listAccessKeys()).length;
-}
+  it('start periodically enforces access key quotas', async (done) => {
```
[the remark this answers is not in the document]
fortuna: Please don't kill the `Clock` interface. Using the native interface makes it hard to work test. It's too magical, and magical is bad. It's a lot easier to understand the Clock interface.

We are "fast forwarding" we do that with the clock.runCallbacks(). Maybe we need to add a way to set the "now" if you wanna test in more details.

src/shadowbox/server/server_access_key.ts
```diff
@@ -102,15 +102,15 @@ export class ServerAccessKeyRepository implements AccessKeyRepository {
 
   // Exposes the access key configuration to the server. Periodically enforces access key quotas.
```
fortuna: Clarify this starts the shadowsocks service?

src/shadowbox/server/server_access_key.ts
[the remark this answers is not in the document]
fortuna: This is not really initialization only. It's actually starting the service.
Maybe `startService`? `startShadowsocksService`? I'm fine with `start` though.

src/shadowbox/model/shadowsocks_server.ts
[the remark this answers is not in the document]
fortuna: I'm fine either way. I feel `Outline` is implied.

src/shadowbox/server/server_access_key.ts
fortuna: Ugh. I feel like we should change the API so that the mutations are methods on the AccessKey object. But that's a change for another PR.

src/shadowbox/server/server_access_key.ts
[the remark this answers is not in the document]
JonathanDCohen: Sure, but when would that happen here? These are all of type AccessKey

src/shadowbox/model/access_key.ts
fortuna: We need to be able to differentiate overridden quota from default quota, even if they are the same. This differentiation will be needed for serialization, but probably also the UI.

I suggest we use the absence of the field to mean default quota, which, by default, is unlimited access. You just need to update the comment.

src/shadowbox/model/access_key.ts
fortuna: Ok, I think the smart unit selection is fine for now. It's easy to extend in a way the user can express the unit.

Not necessary, but maybe we should consider a more extensible format, such as
```
AccessKeyQuota {
  quota: { bytes: number }
  window: { hours: number}
}
```
This is what we use for the [server monthly cost](https://github.com/Jigsaw-Code/outline-server/blob/363c5f16d0019a80981eee11f741c455444f91bc/src/server_manager/model/server.ts#L100)

That can later be extended in a backward-compatible way to something like:
```
AccessKeyQuota {
  quota: { bytes: number, displayUnit: GIGABYTE|MEGABYTE|...}
  window: { hours: number, displayUnit: WEEK|DAY|HOUR}
}
```

src/shadowbox/README.md
fortuna: There's plan to have a default quota, but I agree with Alberto that it belongs on the server level.

src/shadowbox/server/server_access_key.ts
alalamav: Done in 2a653b4.

src/shadowbox/server/manager_metrics.ts
alalamav: It looked like the right place to me, since it already reports outbound metrics, but I see your point in that it serves manager-facing metrics. Implemented an injectable abstraction `UsageMetrics` in 2a653b4.

src/shadowbox/server/server_access_key.ts
alalamav: Done in 2a653b4.

src/shadowbox/server/server_access_key.ts
```diff
@@ -90,73 +123,162 @@ export class ServerAccessKeyRepository implements AccessKeyRepository {
...
-    for (let ai = 0; ai < this.keyConfig.data().accessKeys.length; ai++) {
-      const accessKey = this.keyConfig.data().accessKeys[ai];
+    for (let ai = 0; ai < this.accessKeys.length; ai++) {
+      const accessKey = this.accessKeys[ai];
       if (accessKey.id === id) {
-        this.portProvider.freePort(accessKey.port);
-        this.keyConfig.data().accessKeys.splice(ai, 1);
-        this.keyConfig.write();
+        this.portProvider.freePort(accessKey.proxyParams.portNumber);
+        this.accessKeys.splice(ai, 1);
```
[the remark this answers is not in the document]
fortuna: Talked offline. Shadowsocks doesn't really provide a way to inform the client of a deleted key.

src/shadowbox/server/server_access_key.ts
fortuna: It may be useful to trigger the debug log only in case the state changes, to reduce noise.

src/shadowbox/server/server_access_key.ts
```diff
@@ -90,75 +139,178 @@ export class ServerAccessKeyRepository implements AccessKeyRepository {
...
+      if (accessKey.id === id) {
+        return accessKey;
       }
     }
     return undefined;
   }
 }
+
+// Retrieves access key usage metrics from a Prometheus instance.
+export class AccessKeyUsageMetrics implements UsageMetrics {
```
fortuna: 'Usage', 'ByteTransfer', 'Quota Traffic'. Can we use a consistent phrase?

Maybe `getQuotaTrafficBytes(...)` and `AccessKeyQuotaTraffic implements QuotaTraffic`?

src/shadowbox/model/access_key.ts
alalamav: I think your suggestion is better than hardcoding the unit in the property. Done in a233e1e.

src/shadowbox/server/server_access_key.ts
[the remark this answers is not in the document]
alalamav: As discussed offline, this method was returning a shallow copy. Made the AccessKey interfaces' properties readonly in 2a653b4. Now we just return a copy to the array (of immutable keys).

src/shadowbox/server/server_access_key.ts
alalamav: Done in 20d622d. Went with 'byte transfer' for descriptiveness. This is consistent with the quota feature documentation, as well as with other metrics-related terminology.

src/shadowbox/server/server_access_key.spec.ts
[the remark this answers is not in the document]
JonathanDCohen: Discussed offline, vini and I discussed an implementation like [this](http://www.typescriptlang.org/play/#code/MYGwhgzhAECyYDsCuYQGEQHtgGtoG8AoaE6ABwCcBLANzABcBTaYVEAIzFxgF5oBtALrRIBViBAAuaADEkCYPSqYEAGmhKAto2nJN7RhXVUETCnSnQ9Bo9CoATXUn2GAvkIDcxUpVoNmWjpWzjbQfAAMXqTk1HRMVowAHvQAkvZh0JGE3iQImADuABQAlE4uFATQFIz0SBQIGgAWVBAAdIEe0K7Z0b5xzBCYFPRobJzcJQQ50U0treIcXDi8s20L48vTM6Stg8OFhcAAjOrAAEzFYQB8LEftVNrQALQsZ-faxVGk3dMQNSmmQwWQqBWAAZTKNnUADNpHIFEoVKVguUptsSPRmmsxks2mQkBBGoV8AtpND1IFpJi5oFoABqDQPRjg4yA8yoKlMll2RyrVoIJKpex0umuT5bUjVWr1PkC5JpZ7QI5fEg-aJ-egAFSZmCQ9BBXIhKKh0FhsnkimUCGR1kMaPR1OxEg2eIJRJJbDJFKZVKx72YDNBYNZZgs0gB0JMVHoAE9jLzHfzBWkRWKVTMpXUGom5ULFcrpmrSEpcAbtODIYZLtIaJgHPbtlRodAy8ywdAADwRS6YigFaAAJUQAHNGABRCh9iiFABE2twjMemgJ9GgBnImAg0dojBn4vRdmbIL961xrRAjAQw8xYW7DYPfNpdL4QfTB8z9Tf3x66OAKggq6aGAiTao8L5+k+i5tl4EoYn6ewjDiEz7uiF6rrmGSJqe3D8OEggwQe+TNBeLa5v6nZ8EBIFMpcRAPsWEFMhkZEdLBMxkQsJRftsLFMc+CTJK0Jihqg3EzImCGjM6uJcWx0QYeBczYcsuH4WxRbbImtKUcBoGMOm3Q-GhLDMYw+RwIgKDoFgpbisAuz-GywIAKzqJMPA3H+CCDBe56YMOhTOcUdkOakTmoIUACcbmXB5LD-pgvlYAFkXBV49kanpur6mcADMMXXPF3mJYwfkBXlaWEBlNRZXqgUnC2sWeQlSX+fVwVVfcpbOeEIWZTqdWuY1hVeT5pXJYFQUhSWOCFLl01UKW80eEAA) to allow fast forwarding

src/shadowbox/model/access_key.ts
```diff
@@ -27,12 +27,12 @@ export interface ProxyParams {
   readonly password: string;
 }
 
-// Parameters needed to enforce an access key data quota, over a sliding window.
+// Parameters needed to enforce an access key data transfer quota, over a sliding window.
 export interface AccessKeyQuota {
-  // The allowed metered traffic measured in bytes.
-  readonly quotaBytes: number;
+  // The allowed metered data transfer measured in bytes.
+  readonly quota: {bytes: number};
```
fortuna: Should we call it `data`, `allowedData` or `dataLimit` to avoid the `quota` stuttering?

Having "allowed" or "limit" also makes it super clear it's not the used data.

src/shadowbox/server/server_access_key.spec.ts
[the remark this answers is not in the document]
fortuna: Jon's manual Clock is nice, and it would be good to add to our codebase and use it in the test. But I'd be fine doing that in another PR, to get this big one out of the way and unblock the other changes we want to do.

src/shadowbox/model/access_key.ts
```diff
@@ -18,25 +18,37 @@ export type AccessKeyMetricsId = string;
...
   // Admin-controlled, editable name for this access key.
-  name: string;
+  readonly name: string;
   // Used in metrics reporting to decouple from the real id. Can change.
-  metricsId: AccessKeyMetricsId;
+  readonly metricsId: AccessKeyMetricsId;
   // Parameters to access the proxy
-  proxyParams: ProxyParams;
+  readonly proxyParams: ProxyParams;
+  // Admin-controlled, data transfer quota for this access key. Unlimited if unset.
```
fortuna: Replace "Ulimited if unset" with "Server's default quota (unlimited by default) if unset"?

alalamav: I'll update the comment and other documentation when I implement default quotas.

src/shadowbox/server/server_access_key.spec.ts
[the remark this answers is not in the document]
alalamav: Cool - I'd rather do it in a follow-up PR.

src/shadowbox/model/access_key.ts
alalamav: Done in 93cc9c9.

alalamav:
93cc9c9 updates the model to include access key usage within quotas; this is necessary to report quota usage to the manager. PTAL.

src/shadowbox/server/server_access_key.ts
```diff
@@ -53,8 +53,14 @@ class ServerAccessKey implements AccessKey {
   name: string;
   metricsId: AccessKeyMetricsId;
   readonly proxyParams: ProxyParams;
-  quota?: AccessKeyQuota;
-  isOverQuota?: boolean;
+  quotaUsage?: AccessKeyQuotaUsage;
+}
+
+export function IsAccessKeyOverQuota(accessKey: AccessKey) {
```
fortuna: You could make this an `AccessKey.IsOverQuota()` method

alalamav: Done in 24473f3.

src/shadowbox/server/server_access_key.spec.ts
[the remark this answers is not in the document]
JonathanDCohen: Understood.  Could you throw a quick comment in the test to explain?

src/shadowbox/server/server_access_key.ts
```diff
@@ -187,15 +182,14 @@ export class ServerAccessKeyRepository implements AccessKeyRepository {
...
     }
     const accessKey = this.getAccessKey(id);
     if (!accessKey) {
       return false;
     }
-    accessKey.quota = quota;
+    accessKey.quotaUsage = {quota, usage: {bytes: 0}};
     try {
       this.saveAccessKeys();
       const quotaStautsChanged = await this.updateAccessKeyQuotaStatus(accessKey);
```
JonathanDCohen: This is actually about the catch but github won't let me comment on that line.

Do we want to add some debug logging of the error or some similar thing?  It could be frustrating if an admin can't reset quotas and there's no log of the error.

src/shadowbox/server/server_access_key.ts
```diff
@@ -214,9 +208,8 @@ export class ServerAccessKeyRepository implements AccessKeyRepository {
     if (!accessKey) {
       return false;
     }
-    const wasOverQuota = accessKey.isOverQuota;
-    accessKey.quota = undefined;
-    accessKey.isOverQuota = false;
+    const wasOverQuota = accessKey.isOverQuota();
+    accessKey.quotaUsage = undefined;
```
JonathanDCohen: Same for the catch clause below as well

src/shadowbox/server/server_access_key.ts
alalamav: I think this is part of a larger issue that is the lack of typed errors in this class. Instead of adding logging, which wouldn't be propagated to the API call anyway, I propose we implement typed errors so that the caller can log/communicate the error accordingly.

alalamav:
> Understood. Could you throw a quick comment in the test to explain?

Done in dcc9b16.

src/shadowbox/server/server_access_key.ts
alalamav: See comment above.

src/shadowbox/server/server_access_key.ts
JonathanDCohen: SGTM. can we add a quick todo?

JonathanDCohen: discussed offline.  all good.

JonathanDCohen:
My fault accidentally requesting re-review.  LGTM ^_^

alalamav:
Thanks for the review and comments, @fortuna @JonathanDCohen!

src/shadowbox/server/manager_service.ts
```diff
@@ -168,6 +172,47 @@ export class ShadowsocksManagerService {
...
   }
 
+  public async setAccessKeyQuota(req: RequestType, res: ResponseType, next: restify.Next) {
+    try {
+      logging.debug(`setAccessKeyQuota request ${JSON.stringify(req.params)}`);
+      const accessKeyId = req.params.id;
+      const quota = req.params.quota;
+      // TODO(alalama): remove these checks once the repository supports typed errors.
+      if (!quota || !quota.data || !quota.window) {
+        return next(new restify.InvalidArgumentError(
```
fortuna: I just realized that `restify.InvalidArgumentError` is returning HTTP error 409 Conflict.
That doesn't seem like a standard way of doing that.
Could you change to BadRequest (400) instead?

PR #460, opened 2019-07-23
manager_service cleanup

JonathanDCohen:
Changed magic number http codes to named enum values
De-nested control flow in setShareMetrics()
Formatted.

### DIFF
--- a/src/shadowbox/server/manager_service.ts
+++ b/src/shadowbox/server/manager_service.ts
@@ -60,6 +60,11 @@ interface ResponseType {
   send(code: number, data?: {}): void;
 }
 
+enum Success {
+  OK = 200,
+  NO_CONTENT = 204,
+}
+
 export function bindService(
     apiServer: restify.Server, apiPrefix: string, service: ShadowsocksManagerService) {
   apiServer.put(`${apiPrefix}/name`, service.renameServer.bind(service));
@@ -93,18 +98,18 @@ export class ShadowsocksManagerService {
   public renameServer(req: RequestType, res: ResponseType, next: restify.Next): void {
     const name = req.params.name;
     if (typeof name !== 'string' || name.length > 100) {
-      res.send(400);
-      next();
+      next(new restify.BadRequestError(
+          `Requested server name should be a string <= 100 characters long.  Got ${name}`));
       return;
     }
     this.serverConfig.data().name = name;
     this.serverConfig.write();
-    res.send(204);
+    res.send(Success.NO_CONTENT);
     next();
   }
 
   public getServer(req: RequestType, res: ResponseType, next: restify.Next): void {
-    res.send(200, {
+    res.send(Success.OK, {
       name: this.serverConfig.data().name || this.defaultServerName,
       serverId: this.serverConfig.data().serverId,
       metricsEnabled: this.serverConfig.data().metricsEnabled || false,
@@ -122,7 +127,7 @@ export class ShadowsocksManagerService {
       response.accessKeys.push(accessKeyToJson(accessKey));
     }
     logging.debug(`listAccessKeys response ${response}`);
-    res.send(200, response);
+    res.send(Success.OK, response);
     return next();
   }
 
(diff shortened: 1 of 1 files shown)


src/shadowbox/server/manager_service.ts
```diff
@@ -224,22 +229,22 @@ export class ShadowsocksManagerService {
...
+    }
+    if (req.params.metricsEnabled) {
+      this.metricsPublisher.startSharing();
     } else {
-      res.send(400);
+      this.metricsPublisher.stopSharing();
     }
+    res.send(Success.NO_CONTENT);
     next();
+    // blahblah
```
alalamav: nit: remove

src/shadowbox/server/manager_service.ts
```diff
@@ -224,22 +229,22 @@ export class ShadowsocksManagerService {
...
   public setShareMetrics(req: RequestType, res: ResponseType, next: restify.Next): void {
-    const params = req.params as SetShareMetricsParams;
-    if (typeof params.metricsEnabled === 'boolean') {
-      if (params.metricsEnabled) {
-        this.metricsPublisher.startSharing();
-      } else {
-        this.metricsPublisher.stopSharing();
-      }
-      res.send(204);
+    if (!req.params || typeof req.params.metricsEnabled !== 'boolean') {
```
alalamav: I would leave in the `params` variable with the cast. It makes it explicit that you're dealing with a `SetShareMetricsParam` object and you don't have to keep accessing the request.

src/shadowbox/server/manager_service.ts
```diff
@@ -93,18 +98,18 @@ export class ShadowsocksManagerService {
   public renameServer(req: RequestType, res: ResponseType, next: restify.Next): void {
     const name = req.params.name;
     if (typeof name !== 'string' || name.length > 100) {
-      res.send(400);
-      next();
+      next(new restify.BadRequestError(
+          `Requested server name should be a string <= 100 characters long.  Got ${name}`));
```
alalamav: I wonder if this should be an InvalidArgumentError (409) since the argument itself is incorrect.

src/shadowbox/server/manager_service.ts
```diff
@@ -60,6 +60,11 @@ interface ResponseType {
   send(code: number, data?: {}): void;
 }
 
+enum Success {
```
alalamav: Consider naming `HttpSuccess`, `RestSuccess`, or `SuccessCodes` to further qualify.

src/shadowbox/server/manager_service.ts
```diff
@@ -224,22 +229,22 @@ export class ShadowsocksManagerService {
...
-    if (typeof params.metricsEnabled === 'boolean') {
-      if (params.metricsEnabled) {
-        this.metricsPublisher.startSharing();
-      } else {
-        this.metricsPublisher.stopSharing();
-      }
-      res.send(204);
+    if (!req.params || typeof req.params.metricsEnabled !== 'boolean') {
+      next(new restify.BadRequestError(
+          `Expected metricsEnabled to be boolean.  Instead got ${req.params.metricsEnabled}`));
```
alalamav: The interpolated value may throw if `req` is undefined.  I suggest not echoing it and saying something like  'Must provide a boolean value for metricsEnabled'.

src/shadowbox/server/manager_service.ts
JonathanDCohen: Done.

src/shadowbox/server/manager_service.ts

src/shadowbox/server/manager_service.ts
JonathanDCohen: Sure.  the reason i wanted to get rid of the cast originally was because it seemed to imply that there's some type safety gained there, but it doesn't do any runtime type checks or anything like that.

src/shadowbox/server/manager_service.ts
JonathanDCohen: I split it into two error cases so we can keep the richer error message.

src/shadowbox/server/manager_service.ts

PR #467, opened 2019-08-16
fix DigitalOcean API not allow CORS

jinrenjie:
fix DigitalOcean API not allow CORS


alalamav:
Thanks for the PR @jinrenjie. We have fixed the issue without disabling SOP in #468.

PR #485, opened 2019-09-11
Server data usage timeframe

alalamav:
* Implements a server data usage timeframe, used to report access key data usage metrics (`GET /metrics/transfer`) and to enforce access key data limits.
* Exposes an API endpoint, `PUT /server/data-usage-timeframe`, to configure this value. The default is 30 days for backwards compatibility with the existing `GET metrics/transfer` timeframe.
* The data usage timeframe can be retrieved through the `GET /server` endpoint.
* Previously, access key data limits supported individual timeframes per key in order to calculate data utlization. The timeframe is now shared between all access keys with data limits.
* Updates the `PUT /access-key/[:id]/data-limit` API to reflect the shared timeframe.

### DIFF
--- a/src/shadowbox/infrastructure/prometheus_scraper.ts
+++ b/src/shadowbox/infrastructure/prometheus_scraper.ts
@@ -99,9 +99,7 @@ export function runPrometheusScraper(
 async function waitForPrometheusReady(prometheusEndpoint: string) {
   logging.debug('Waiting for Prometheus to be ready...');
   while (!(await isHttpEndpointHealthy(prometheusEndpoint))) {
-    await new Promise((resolve) => setTimeout(() => {
-                        resolve();
-                      }, 1000));
+    await new Promise((resolve) => setTimeout(resolve, 1000));
   }
   logging.debug('Prometheus is ready');
 }
--- a/src/shadowbox/infrastructure/prometheus_scraper.ts
+++ b/src/shadowbox/infrastructure/prometheus_scraper.ts
@@ -69,31 +69,38 @@ export class PrometheusClient {
   }
 }
 
-export function runPrometheusScraper(
+export async function runPrometheusScraper(
     args: string[], configFilename: string, configJson: {},
     prometheusEndpoint: string): Promise<child_process.ChildProcess> {
   mkdirp.sync(path.dirname(configFilename));
   const ymlTxt = jsyaml.safeDump(configJson, {'sortKeys': true});
-  fs.writeFileSync(configFilename, ymlTxt, 'utf-8');
-  return new Promise(async (resolve, reject) => {
-    const commandArguments = ['--config.file', configFilename];
-    commandArguments.push(...args);
-    const runProcess = child_process.spawn('/root/shadowbox/bin/prometheus', commandArguments);
-    runProcess.on('error', (error) => {
-      logging.error(`Error spawning prometheus: ${error}`);
+  // Write the file asynchronously to prevent blocking the node thread.
+  await new Promise((resolve, reject) => {
+    fs.writeFile(configFilename, ymlTxt, 'utf-8', (err) => {
+      if (err) {
+        reject(err);
+      } else {
+        resolve();
+      }
     });
-    // TODO(fortuna): Add restart logic.
-    runProcess.on('exit', (code, signal) => {
-      logging.info(`prometheus has exited with error. Code: ${code}, Signal: ${signal}`);
-    });
-    // TODO(fortuna): Disable this for production.
-    // TODO(fortuna): Consider saving the output and expose it through the manager service.
-    runProcess.stdout.pipe(process.stdout);
-    runProcess.stderr.pipe(process.stderr);
-
-    await waitForPrometheusReady(prometheusEndpoint);
-    resolve(runProcess);
   });
+  const commandArguments = ['--config.file', configFilename];
+  commandArguments.push(...args);
+  const runProcess = child_process.spawn('/root/shadowbox/bin/prometheus', commandArguments);
+  runProcess.on('error', (error) => {
+    logging.error(`Error spawning prometheus: ${error}`);
+  });
+  // TODO(fortuna): Add restart logic.
+  runProcess.on('exit', (code, signal) => {
+    logging.info(`prometheus has exited with error. Code: ${code}, Signal: ${signal}`);
+  });
+  // TODO(fortuna): Disable this for production.
+  // TODO(fortuna): Consider saving the output and expose it through the manager service.
+  runProcess.stdout.pipe(process.stdout);
+  runProcess.stderr.pipe(process.stderr);
+
+  await waitForPrometheusReady(prometheusEndpoint);
+  return runProcess;
 }
 
 async function waitForPrometheusReady(prometheusEndpoint: string) {
--- a/src/shadowbox/infrastructure/prometheus_scraper.ts
+++ b/src/shadowbox/infrastructure/prometheus_scraper.ts
@@ -74,44 +74,41 @@ export function runPrometheusScraper(
     prometheusEndpoint: string): Promise<child_process.ChildProcess> {
   mkdirp.sync(path.dirname(configFilename));
   const ymlTxt = jsyaml.safeDump(configJson, {'sortKeys': true});
-  return new Promise((resolve, reject) => {
-    fs.writeFile(configFilename, ymlTxt, 'utf-8', async (err) => {
-      if (err) {
-        reject(err);
-      }
-      const commandArguments = ['--config.file', configFilename];
-      commandArguments.push(...args);
-      const runProcess = child_process.spawn('/root/shadowbox/bin/prometheus', commandArguments);
-      runProcess.on('error', (error) => {
-        logging.error(`Error spawning prometheus: ${error}`);
-      });
-      // TODO(fortuna): Add restart logic.
-      runProcess.on('exit', (code, signal) => {
-        logging.info(`prometheus has exited with error. Code: ${code}, Signal: ${signal}`);
-      });
-      // TODO(fortuna): Disable this for production.
-      // TODO(fortuna): Consider saving the output and expose it through the manager service.
-      runProcess.stdout.pipe(process.stdout);
-      runProcess.stderr.pipe(process.stderr);
-
-      await waitForPrometheusReady(prometheusEndpoint);
-      resolve(runProcess);
+  fs.writeFileSync(configFilename, ymlTxt, 'utf-8');
+  return new Promise(async (resolve, reject) => {
+    const commandArguments = ['--config.file', configFilename];
+    commandArguments.push(...args);
+    const runProcess = child_process.spawn('/root/shadowbox/bin/prometheus', commandArguments);
+    runProcess.on('error', (error) => {
+      logging.error(`Error spawning prometheus: ${error}`);
     });
+    // TODO(fortuna): Add restart logic.
+    runProcess.on('exit', (code, signal) => {
+      logging.info(`prometheus has exited with error. Code: ${code}, Signal: ${signal}`);
+    });
+    // TODO(fortuna): Disable this for production.
+    // TODO(fortuna): Consider saving the output and expose it through the manager service.
+    runProcess.stdout.pipe(process.stdout);
+    runProcess.stderr.pipe(process.stderr);
+
+    await waitForPrometheusReady(prometheusEndpoint);
+    resolve(runProcess);
   });
 }
 
 async function waitForPrometheusReady(prometheusEndpoint: string) {
   logging.debug('Waiting for Prometheus to be ready...');
-  let ready = false;
-  while (!ready) {
-    ready = await isPrometheusReady(prometheusEndpoint);
+  while (!(await isHttpEndpointHealthy(prometheusEndpoint))) {
+    await new Promise((resolve) => setTimeout(() => {
+                        resolve();
+                      }, 1000));
   }
   logging.debug('Prometheus is ready');
 }
 
-function isPrometheusReady(prometheusEndpoint: string): Promise<boolean> {
+function isHttpEndpointHealthy(endpoint: string): Promise<boolean> {
   return new Promise((resolve, reject) => {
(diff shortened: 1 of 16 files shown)


src/shadowbox/server/manager_service.ts
```diff
@@ -258,9 +265,36 @@ export class ShadowsocksManagerService {
...
+  public setDataUsageTimeframe(req: RequestType, res: ResponseType, next: restify.Next) {
+    try {
+      logging.debug(`setDataUsageTimeframe request ${JSON.stringify(req.params)}`);
+      const hours = req.params.hours;
+      if (hours === undefined) {  // The access key repository will validate the value.
+        return next(
+            new restify.MissingParameterError({statusCode: 400}, 'Missing `hours` parameter'));
+      } else if (!Number.isInteger(hours)) {
+        return next(
+            new restify.InvalidArgumentError({statusCode: 400}, '`hours` must be an integer'));
```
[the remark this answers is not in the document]
alalamav: As far as I understand, Restify doesn't perform parameter validation, so I think the check is reachable.

src/shadowbox/server/manager_metrics.ts
```diff
@@ -13,23 +13,23 @@
...
+  getOutboundByteTransfer(timeframe: DataUsageTimeframe): Promise<DataUsageByUser>;
 }
 
 // Reads manager metrics from a Prometheus instance.
 export class PrometheusManagerMetrics implements ManagerMetrics {
   constructor(private prometheusClient: PrometheusClient) {}
 
-  async get30DayByteTransfer(): Promise<DataUsageByUser> {
+  async getOutboundByteTransfer(timeframe: DataUsageTimeframe): Promise<DataUsageByUser> {
     // TODO(fortuna): Consider pre-computing this to save server's CPU.
```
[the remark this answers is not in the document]
alalamav: We populate `AccessKey.dataUsage` through `getOutboundByteTransfer`. I removed `ServerAccessKeyRepository.getDataUsage` because, with the shared timeframe, there is no need to compute individual access key usage.

src/shadowbox/server/manager_service.ts
```diff
@@ -52,8 +52,9 @@ interface RequestParams {
   id?: string;
   name?: string;
   metricsEnabled?: boolean;
-  limit?: AccessKeyDataLimit;
+  limit?: DataUsage;
```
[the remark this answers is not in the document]
alalamav: Done. Other fields are primitives and already marked as potentially undefined.

src/shadowbox/server/api.yml
```diff
@@ -292,18 +314,19 @@ components:
...
           type: integer
-    AccessKeyDataLimit:
+        dataUsageTimeframe:
+          $ref: "#/components/schemas/DataUsageTimeframe"
+
+    DataUsageTimeframe:
+      properties:
+        hours:
+          type: integer
+
```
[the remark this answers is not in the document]
alalamav: Added minimum values to the API documentation.

src/shadowbox/server/manager_service.ts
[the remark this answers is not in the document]
bemasc: OK.  Is api.yml just for documentation?  Does anything process it at all?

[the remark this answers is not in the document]
alalamav: I believe it's documentation only.

src/shadowbox/server/server_access_key.ts
```diff
@@ -44,22 +46,21 @@ export interface AccessKeyConfigJson {
...
-    return this.dataLimitUsage.usage.bytes > this.dataLimitUsage.limit.data.bytes;
+    return this.dataUsage.bytes > this.dataLimit.bytes;
   }
 }
 
-function isValidAccessKeyDataLimit(limit: AccessKeyDataLimit) {
-  return limit && limit.data && limit.timeframe && Number.isInteger(limit.data.bytes) &&
-      limit.data.bytes >= 0 && Number.isInteger(limit.timeframe.hours) &&
-      limit.timeframe.hours >= 0;
+function isValidAccessKeyDataLimit(limit: DataUsage) {
```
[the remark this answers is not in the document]
alalamav: tsc can infer types, but added the type for readability.

src/shadowbox/server/manager_service.ts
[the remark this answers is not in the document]
bemasc: OK.  Note that the other fields remain unsafe: we don't actually know that they are primitives or undefined.  The caller could instead pass us an object, a list, or the wrong kind of primitive, and the typescript typing would be incorrect.

src/shadowbox/server/mocks/mocks.ts
```diff
@@ -56,15 +56,8 @@ export class FakePrometheusClient extends PrometheusClient {
...
-    let accessKeyIds: string[];
-    if (accessKeyIdMatch && accessKeyIdMatch[1]) {
-      // Return query results only for the specified access key.
-      accessKeyIds = [accessKeyIdMatch[1]];
-    } else {
-      accessKeyIds = Object.keys(this.bytesTransferredById);
-    }
     const queryResultData = {result: []} as QueryResultData;
-    for (const accessKeyId of accessKeyIds) {
+    for (const accessKeyId of Object.keys(this.bytesTransferredById)) {
```
bemasc: Can you use `for(const accessKeyId in this.bytesTransferredById)` instead?

alalamav: Technically yes, but the compiler complains with: `for (... in ...) statements must be filtered with an if statement`. `Object.keys` avoids the check.

src/shadowbox/server/manager_metrics.ts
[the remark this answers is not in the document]
bemasc: OK, but `ShadowsocksManagerService.getDataUsage` still calls `this.managerMetrics.getOutboundByteTransfer`, which is a full recomputation from Prometheus.  Now that we have hourly recomputations for all access keys, `getDataUsage` could return those cached values instead of forcing a recalculation.

We don't have to do this now, but I think the TODO here has become confusing because half of it is now implemented, and the other half wouldn't be implemented at this point in the code.

src/shadowbox/server/mocks/mocks.ts
bemasc: Yuck.  OK.

src/shadowbox/server/manager_service.ts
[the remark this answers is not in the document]
fortuna: Restify does not know the expected types. It will only parse the JSON. 
The .yml schema is currently only used for documentation

src/shadowbox/server/manager_metrics.ts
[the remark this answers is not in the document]
fortuna: Oh I like that. Here is a possible design:

- PrometheusManagerMetrics runs the polling loop to update the usage. We should do it more often. Maybe once a minute? What do we use in the manager polling? We would need to check the CPU.
- The bytes transferred API doesn't hit prometheus. It only returns the currently cached values.
- PrometheusManagerMetrics exposes a "metric update" event others can subscribe
- We use the event to enforce the data limits
- We use the event in the data limit tests tests (which can actually inject a fake ManagerMetrics).

This way:
- the quota enforcement is really driven by how fast we update the usage
- The manager can hit the api as often as it wants and it won't cause any more load.
- Later we may even change the manager to an event-based API that keeps an HTTPS connection waiting for the next metric update.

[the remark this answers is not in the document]
fortuna: BTW, the TODO was about materializing the query on Prometheus, but it's effectively the same.
We can't actually do it on Prometheus now though, since the timeframe can change (it was fixed before)

src/shadowbox/infrastructure/prometheus_scraper.ts
```diff
@@ -70,11 +70,12 @@ export class PrometheusClient {
...
 
 export function runPrometheusScraper(
-    args: string[], configFilename: string, configJson: {}): Promise<child_process.ChildProcess> {
+    args: string[], configFilename: string, configJson: {},
+    prometheusEndpoint: string): Promise<child_process.ChildProcess> {
   mkdirp.sync(path.dirname(configFilename));
   const ymlTxt = jsyaml.safeDump(configJson, {'sortKeys': true});
   return new Promise((resolve, reject) => {
-    fs.writeFile(configFilename, ymlTxt, 'utf-8', (err) => {
+    fs.writeFile(configFilename, ymlTxt, 'utf-8', async (err) => {
```
fortuna: This is fine, but I realized you can unnest the logic if you prefer (optional):
```ts
export async function runPrometheusScraper(..) {
...
// Async'fy writeFile
await new Promise((resolve, reject) => {
  fs.writeFile(configFilename, ymlTxt, 'utf-8', (err) => {
    if (err) { reject(err); }
    else { resolve(); }
  });
})
const commandArguments = ['--config.file', configFilename];
... // existing code here
runProcess.stderr.pipe(process.stderr);
await waitForPrometheusReady(prometheusEndpoint)
return process
```

src/shadowbox/infrastructure/prometheus_scraper.ts
```diff
@@ -92,7 +93,29 @@ export function runPrometheusScraper(
...
     });
   });
 }
+
+async function waitForPrometheusReady(prometheusEndpoint: string) {
+  logging.debug('Waiting for Prometheus to be ready...');
+  let ready = false;
+  while (!ready) {
+    ready = await isPrometheusReady(prometheusEndpoint);
+  }
```
fortuna: This may flood the server with requests. Maybe sleep for a second between attempts?

src/shadowbox/infrastructure/prometheus_scraper.ts
```diff
@@ -92,7 +93,29 @@ export function runPrometheusScraper(
...
+  logging.debug('Prometheus is ready');
+}
+
+function isPrometheusReady(prometheusEndpoint: string): Promise<boolean> {
```
fortuna: This is not Prometheus-specific. Maybe rename to something like "isHttpEndpointHealthy"?

src/shadowbox/infrastructure/prometheus_scraper.ts
alalamav: Good suggestion, I'm using `fs.writeFileSync` instead of waiting for the async version.

src/shadowbox/infrastructure/prometheus_scraper.ts
alalamav: Done.

src/shadowbox/infrastructure/prometheus_scraper.ts
alalamav: Done.

src/shadowbox/infrastructure/prometheus_scraper.ts
```diff
@@ -74,44 +74,41 @@ export function runPrometheusScraper(
...
-        logging.info(`prometheus has exited with error. Code: ${code}, Signal: ${signal}`);
-      });
-      // TODO(fortuna): Disable this for production.
-      // TODO(fortuna): Consider saving the output and expose it through the manager service.
-      runProcess.stdout.pipe(process.stdout);
-      runProcess.stderr.pipe(process.stderr);
-
-      await waitForPrometheusReady(prometheusEndpoint);
-      resolve(runProcess);
+  fs.writeFileSync(configFilename, ymlTxt, 'utf-8');
```
fortuna: Please revert to the async write. The problem with this one is that it will block the node thread. It's not equivalent to the await with promise, since await doesn't block the thread.

src/shadowbox/infrastructure/prometheus_scraper.ts
```diff
@@ -74,44 +74,41 @@ export function runPrometheusScraper(
...
+  return new Promise(async (resolve, reject) => {
```
fortuna: You don't need this Promise if you make runPrometheusScraper async

src/shadowbox/infrastructure/prometheus_scraper.ts
```diff
@@ -74,44 +74,41 @@ export function runPrometheusScraper(
...
+    // TODO(fortuna): Add restart logic.
+    runProcess.on('exit', (code, signal) => {
+      logging.info(`prometheus has exited with error. Code: ${code}, Signal: ${signal}`);
     });
+    // TODO(fortuna): Disable this for production.
+    // TODO(fortuna): Consider saving the output and expose it through the manager service.
+    runProcess.stdout.pipe(process.stdout);
+    runProcess.stderr.pipe(process.stderr);
+
+    await waitForPrometheusReady(prometheusEndpoint);
```
fortuna: All the code above could be moved out of the Promise.

alalamav: Done.

src/shadowbox/infrastructure/prometheus_scraper.ts
alalamav: Done.

src/shadowbox/infrastructure/prometheus_scraper.ts
alalamav: Done, added a comment to explain this behavior.

src/shadowbox/infrastructure/prometheus_scraper.ts
```diff
@@ -69,30 +69,55 @@ export class PrometheusClient {
...
+  commandArguments.push(...args);
+  const runProcess = child_process.spawn('/root/shadowbox/bin/prometheus', commandArguments);
+  runProcess.on('error', (error) => {
+    logging.error(`Error spawning prometheus: ${error}`);
+  });
+  // TODO(fortuna): Add restart logic.
+  runProcess.on('exit', (code, signal) => {
+    logging.info(`prometheus has exited with error. Code: ${code}, Signal: ${signal}`);
+  });
+  // TODO(fortuna): Disable this for production.
```
fortuna: I think we can remove this production comment... 🙄

alalamav: Agreed, done.

PR #492, opened 2019-09-16
Improve type safety of manager service request parameters

alalamav:
* Follow up to a review comment of #485.
* Adds a user type guard to the manager API request parameters.
* Changes the type of `RequestType.params` to  `RequestParams|{}` to reflect incoming untyped data.
* `s/errors.InvalidDataLimitTimeframe/errors.InvalidDataUsageTimeframe`

FYI, @fortuna, @JonathanDCohen.

### DIFF
--- a/src/shadowbox/server/manager_service.ts
+++ b/src/shadowbox/server/manager_service.ts
@@ -98,7 +98,7 @@ interface SetShareMetricsParams {
 }
 
 // User type guard to validate that `params` is of type RequestParams and contains `property`.
-function hasParameter(params: RequestParams|{}, property: string): params is RequestParams {
+function isRequestParam(params: RequestParams|{}, property: string): params is RequestParams {
   return property in (params as RequestParams);
 }
 
@@ -113,7 +113,7 @@ export class ShadowsocksManagerService {
 
   public renameServer(req: RequestType, res: ResponseType, next: restify.Next): void {
     logging.debug(`renameServer request ${JSON.stringify(req.params)}`);
-    if (!hasParameter(req.params, 'name')) {
+    if (!isRequestParam(req.params, 'name')) {
       return next(
           new restify.MissingParameterError({statusCode: 400}, 'Parameter `name` is missing'));
     }
@@ -173,7 +173,7 @@ export class ShadowsocksManagerService {
       Promise<void> {
     try {
       logging.debug(`setPortForNewAccessKeys request ${JSON.stringify(req.params)}`);
-      if (!hasParameter(req.params, 'port')) {
+      if (!isRequestParam(req.params, 'port')) {
         return next(
             new restify.MissingParameterError({statusCode: 400}, 'Parameter `port` is missing'));
       }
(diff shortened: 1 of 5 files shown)


src/shadowbox/server/manager_service.ts
```diff
@@ -52,12 +52,12 @@ interface RequestParams {
...
   name?: string;
   metricsEnabled?: boolean;
-  limit?: DataUsage|{};
+  limit?: DataUsage;
   port?: number;
   hours?: number;
 }
 interface RequestType {
-  params: RequestParams;
+  params: RequestParams|{};
```
bemasc: If we make that change to `RequestParams`, then `params` doesn't need the `|{}`.

src/shadowbox/server/manager_service.ts
```diff
@@ -210,6 +223,14 @@ export class ShadowsocksManagerService {
...
+      if (!isRequestParam(req.params, 'id')) {
+        return next(
+            new restify.MissingParameterError({statusCode: 400}, 'Parameter `id` is missing'));
+      }
+      if (!isRequestParam(req.params, 'name')) {
+        return next(
+            new restify.MissingParameterError({statusCode: 400}, 'Parameter `name` is missing'));
+      }
       const accessKeyId = req.params.id;
       this.accessKeys.renameAccessKey(accessKeyId, req.params.name);
```
bemasc: As an example of the kind of thing I'm worried about: even with the new checks, there's nothing here that stops the caller from passing a non-`string` value for `req.params.name`, which could have all kinds of weird effects.

src/shadowbox/server/manager_service.ts
```diff
@@ -52,12 +52,12 @@ interface RequestParams {
   id?: string;
   name?: string;
   metricsEnabled?: boolean;
-  limit?: DataUsage|{};
+  limit?: DataUsage;
```
bemasc: I think the best solution here is to make `RequestParams` be the type `[key: string]: unknown`.  This accurately reflects what we know about `params` when we receive it.

The functions would then look like
```ts
const id = params['id']
if (typeof id !== "string") { return ... }
// id is now a string
```

alalamav:
@bemasc, I followed your suggestion to make `RequestParams` type `[key: string]: unknown`. I think it makes the code more resistent to type errors, though we loose the self-documenting nature of the previous implementation. PTAL.

src/shadowbox/server/manager_service.ts
```diff
@@ -269,12 +297,14 @@ export class ShadowsocksManagerService {
...
-        return next(
-            new restify.MissingParameterError({statusCode: 400}, 'Missing `hours` parameter'));
-      } else if (!Number.isInteger(hours)) {
+      if (!hours) {
         return next(
-            new restify.InvalidArgumentError({statusCode: 400}, '`hours` must be an integer'));
+            new restify.MissingParameterError({statusCode: 400}, 'Parameter `hours` is missing'));
+      }
+      if (typeof hours !== 'number' ||
+          !Number.isInteger(hours)) {  // The access key repository will validate the value.
```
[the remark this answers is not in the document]
alalamav: I believe it still applies, the repository verifies that it's a positive integer.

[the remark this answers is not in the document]
bemasc: OK

PR #496, opened 2019-09-18
[feat]: support build for ARM architectures

oreoluwa:
image: oreoluwa/shadowbox

Just a reference implementation for building for the arm architecture.


oreoluwa:
@JonathanDCohen totally agree with all the comments above, thanks for reviewing.
PS: 
1. this PR was done/raised before the first build of the `outline-ss-server` for the arm arch.
2. it wasn't really meant as a PR, but a reference implementation for anyone wanting to build the outline-server for their arm devices, based on my experience while trying to run outline on my RaspberryPi.
3. I can clean this up in the coming week, with the updated binaries (since prometheus and outline-ss-server now have official arm releases) and incorporate your comments above, if support for arm is currently being considered.

JonathanDCohen:
@oreoluwa We're definitely down for best-effort ARM support although it isn't currently in our CI suite.  I'll take a look whenever you get around to updating :)

andela-oakinniranye:
@JonathanDCohen implemented the multi-arch build. However, I'm not sure I fully understand the CI setup yet, but I'd propose we do a multi-arch deployment, with `docker manifest`, that way the user wouldn't need to specify the tag for the image to be deployed and can just use the `stable` or `daily*` image; seems that's the approach watchtower and Prometheus also took: https://github.com/containrrr/watchtower/blob/master/.circleci/config.yml#L133-L206
https://github.com/prometheus/prometheus/blob/master/Makefile.common#L211-L233
PS: this PR also enables the option to specify the `WATCHTOWER_IMAGE` image, as the default `v2tech/watchtower` has been replaced with `containrrr/watchtower` which has a multi-arch support as well.

PR #504, opened 2019-09-20
Fix unhandled rejection

alalamav:
This can happen after an access key has been deleted.


JonathanDCohen:
looks good.  Do you think it's worthwhile to add a test for getMetricsId()?

alalamav:
We can, but I don't think it's necessary since the method is a getter.

JonathanDCohen:
Yeah I think that's fine.  Approved

PR #505, opened 2019-09-27
Allow the port for new access keys to be editable from the UI.

JonathanDCohen:
Not included in this PR is checking to see if the underlying server supports this feature.  For now it displays "unknown error", this will be implemented as a separate follow-up.


JonathanDCohen:
Friendly ping :)

src/server_manager/messages/en.json
```diff
@@ -105,6 +110,8 @@
   "notification-feedback-thanks": "Thanks for helping us improve! We love hearing from you.",
   "notification-key-added": "Key added",
   "notification-key-removed": "Key removed",
+  "notification-keys-port-saving": "Saving...",
+  "notification-keys-port-saved": "Saved",
```
alalamav: The keys for these could be more general (not specific to keys-port), i.e. `notificiation-saving`, `notification-saved`.

src/server_manager/messages/en.json
```diff
@@ -154,6 +161,8 @@
   "servers-digitalocean": "DigitalOcean servers",
   "servers-manual": "Servers",
   "settings-access-key-port": "Port for new access keys",
+  "settings-access-key-port-save": "Save",
+  "settings-access-key-port-cancel": "Cancel",
```
alalamav: We already have a "cancel" key that you can use. See above for "save".

src/server_manager/messages/en.json
```diff
@@ -40,15 +40,20 @@
   "error-key-remove": "Failed to remove key",
   "error-key-rename": "Failed to rename key",
   "error-keys-get": "Could not load keys",
+  "error-keys-port-bad-input": "The port must be an integer between 1 and 65,535.",
+  "error-keys-port-in-use": "The port is already in use on the server.",
+  "error-keys-port-not-saved": "Not Saved",
```
alalamav: We can make this key more general, i.e. `error-not-saved`

src/server_manager/messages/en.json
```diff
@@ -40,15 +40,20 @@
...
   "error-metrics": "Error setting metrics enabled",
+  "error-network": "A network error occurred.",
   "error-server-creation": "There was an error creating your Outline server.",
   "error-server-destroy": "Failed to destroy server",
   "error-server-removed": "{serverName} no longer present in your DigitalOcean account.",
   "error-server-rename": "Failed to rename server",
   "error-server-unreachable": "Your Outline Server was installed correctly, but we are not able to connect to it. Most likely this is because your server's firewall rules are blocking incoming connections. Please review them and make sure to allow incoming TCP connections on ports ranging from 1024 to 65535.",
   "error-server-unreachable-title": "Unable to connect to your Outline Server",
   "error-servers-removed": "{serverNames} no longer present in your DigitalOcean account.",
+  "error-unknown": "An unknown error occurred.",
```
alalamav: s/unknown/unexpected?

src/server_manager/model/server.ts
```diff
@@ -58,6 +58,9 @@ export interface Server {
   // Returns the port number for new access keys.
   // Returns undefined if the server doesn't have a port set.
   getPortForNewAccessKeys(): number|undefined;
+
+  // Changes the port newly created access keys connect to.
```
alalamav: nit: the wording here can be improved

src/server_manager/ui_components/outline-server-settings.html
```diff
@@ -208,6 +240,76 @@ <h3>[[localize('settings-metrics-header')]]</h3>
...
+        this.$.keysPortCancelButton.disabled = !enabled;
+      },
+      _handleKeysPortClick: function() {
+        this._keysPortButtonsVisible(true);
+        const invalid = this.$.keysPortInput.invalid;
+        this.$.keysPortSaveButton.disabled = invalid;
+        this.$.keysPortCancelButton.disabled = false;
+      },
+      _handleKeysPortServerSuccess() {
+        console.info('Port successfully changed');
```
alalamav: We generally avoid logging in HTML. We should be able to tell whether the port change succeded form the API request logs.

src/server_manager/ui_components/outline-server-settings.html
```diff
@@ -208,6 +240,76 @@ <h3>[[localize('settings-metrics-header')]]</h3>
...
+        this.serverPortForNewAccessKeys = Number(this.$.keysPortInput.value);
+        this._keysPortButtonsVisible(false);
+        this.$.keysPortInput.readonly = false;
+      },
+      _handleKeysPortServerFail(message) {
+        console.error('Port failed to be changed');
```
alalamav: Same here.

src/server_manager/ui_components/outline-server-settings.html
```diff
@@ -208,6 +240,76 @@ <h3>[[localize('settings-metrics-header')]]</h3>
...
+        const input = this.$.keysPortInput;
+        if (Number(input.value) === this.serverPortForNewAccessKeys) {
+          this._handleKeysPortCancel();
+          return;
+        };
+        this._keysPortButtonsEnabled(false);
+        input.readonly = true;
+        input.invalid = false;
+        this.fire('ChangePortForNewAccessKeysRequested', {
+          port: Number(this.$.keysPortInput.value),
```
alalamav: `Number(input.value)`

src/server_manager/ui_components/outline-server-settings.html
```diff
@@ -208,6 +240,76 @@ <h3>[[localize('settings-metrics-header')]]</h3>
           this.fire(metricsSignal);
         }
       },
+      _keysPortButtonsVisible: function(visible) {
```
alalamav: nit: use a verb for methods, i.e. `_togglePortButtonsVisible`

src/server_manager/ui_components/outline-server-settings.html
```diff
@@ -208,6 +240,76 @@ <h3>[[localize('settings-metrics-header')]]</h3>
...
+        this.$.keysPortSaveButton.hidden = !visible;
+        this.$.keysPortCancelButton.hidden = !visible;
+      },
+      _keysPortButtonsEnabled: function(enabled) {
```
alalamav: Same here: `toggleKeysPortButtonsEnabled`

src/server_manager/web_app/app.ts
```diff
@@ -911,6 +915,28 @@ export class App {
         });
   }
 
+  private setPortForNewAccessKeys(port: number, success: () => void, fail: (message: string) => void) {
+    this.appRoot.showNotification(this.appRoot.localize("notification-keys-port-saving"));
```
alalamav: You could rewrite this function with async/await to avoid nesting.

src/server_manager/ui_components/outline-server-settings.html
```diff
@@ -133,7 +161,11 @@ <h3>DigitalOcean</h3>
             <h3>[[localize('settings-server-info')]]</h3>
             <paper-input id="serverNameInput" class="server-name" value="{{name}}" label="[[localize('settings-server-name')]]" always-float-label maxlength="100" on-keydown="_handleNameInputKeyDown" on-blur="_handleNameInputBlur"></paper-input>
             <p class="detail">[[localize('settings-server-rename')]]</p>
-            <paper-input readonly value="[[serverPortForNewAccessKeys]]" label="[[localize('settings-access-key-port')]]" hidden$="[[!serverPortForNewAccessKeys]]" always-float-label maxlength="100"></paper-input>
+            <div id="portForNewAccessKeysContainer">
+              <paper-input id="keysPortInput" value="[[serverPortForNewAccessKeys]]" label="[[localize('settings-access-key-port')]]" maxlength="5" error-color="#ffffff" error-message="ERRORMESSSAGE" hidden$="[[!serverPortForNewAccessKeys]]" on-click="_handleKeysPortClick" on-keyup="_handleKeysPortInputKeyDown" always-float-label maxlength="100"></paper-input>
```
alalamav: Do we want this: `message="ERRORMESSAGE"`?

alalamav: Maybe you can style `error-color` with CSS?

alalamav: We can only allow numeric input on the form by using: `auto-validate allowed-pattern="[0-9]{5}" error-message=[[localize('error-keys-port-bad-input')]]` (double check the regex). Then we can just validate that the port is within range in `_handleKeysPortInputKeyDown`.

We can also use a custom validator (`validator="_validatePortInRange"`) to separate the key stroke handling (esc, enter) and the port range validation.

JonathanDCohen: I chose not to use a custom validator since the validator and the server response would be fighting for control of the invalid property: from https://www.webcomponents.org/element/@polymer/paper-input/elements/paper-input#property-invalid

> If autoValidate is true, the invalid attribute is managed automatically, which can clobber attempts to manage it manually.

So I sidestepped the issue.

JonathanDCohen: It was already set above with `--paper-input-container-invalid-color`.  Removed the extra noise.

JonathanDCohen: It never gets displayed, since the error states all set their own message.  Changed to an empty string

src/server_manager/web_app/app.ts
JonathanDCohen: None of the other event listeners use async/await, and I thought it would be negative for readability to have just one that is, as a future reviewer would wonder why this one is different from the others.  It also only saves nesting only on a couple lines, all the error handling would still have the same indentation level inside the catch block.

src/server_manager/ui_components/outline-server-settings.html
JonathanDCohen: Done

src/server_manager/model/server.ts
JonathanDCohen: Changed to match the getter

src/server_manager/messages/en.json
JonathanDCohen: Done

src/server_manager/messages/en.json
JonathanDCohen: Done

src/server_manager/messages/en.json
JonathanDCohen: Done

JonathanDCohen:
Responded to comments, PTAL :)

src/server_manager/ui_components/outline-server-settings.html
alalamav: That's fine. I still think we'd benefit from only allowing numeric input: `pattern="[0-9]{5}"`.

src/server_manager/web_app/app.ts
alalamav: When this code was written, async/await were not yet available. There is currently a mix of promise and async/await in this file. I think we should eventually migate the old code and not introduce more promise-style code in new PRs.

src/server_manager/ui_components/outline-server-settings.html
alalamav: We actually don't need the `error-message` attribute. Please remove it.

alalamav: nit: style the underline with primary green.

Screen Shot 2019-10-08 at 10 57 34 AM

src/server_manager/messages/en.json
alalamav: I think you missed the error string.

src/server_manager/ui_components/outline-server-settings.html
JonathanDCohen: Done

JonathanDCohen: Done

JonathanDCohen: Done

src/server_manager/messages/en.json
JonathanDCohen: Done

JonathanDCohen:
This is now a feature-complete PR including disabling editing the port in the UI if the server doesn't support it.

src/server_manager/ui_components/outline-server-settings.html
```diff
@@ -133,7 +165,11 @@ <h3>DigitalOcean</h3>
             <h3>[[localize('settings-server-info')]]</h3>
             <paper-input id="serverNameInput" class="server-name" value="{{name}}" label="[[localize('settings-server-name')]]" always-float-label maxlength="100" on-keydown="_handleNameInputKeyDown" on-blur="_handleNameInputBlur"></paper-input>
             <p class="detail">[[localize('settings-server-rename')]]</p>
-            <paper-input readonly value="[[serverPortForNewAccessKeys]]" label="[[localize('settings-access-key-port')]]" hidden$="[[!serverPortForNewAccessKeys]]" always-float-label maxlength="100"></paper-input>
+            <div id="portForNewAccessKeysContainer">
+              <paper-input id="keysPortInput" readonly$="[[!isAccessKeyPortEditable]]" allowed-pattern="[0-9]{1,5}" value="[[serverPortForNewAccessKeys]]" label="[[localize('settings-access-key-port')]]" maxlength="5" hidden$="[[!serverPortForNewAccessKeys]]" on-click="_handleKeysPortClick" on-keyup="_handleKeysPortInputKeyDown" always-float-label maxlength="100"></paper-input>
```
fortuna: We should be using on-tap instead of on-click:
https://stackoverflow.com/questions/26386369/when-to-use-on-click-or-on-tap-w-polymer

src/server_manager/ui_components/outline-server-settings.html
```diff
@@ -208,6 +245,77 @@ <h3>[[localize('settings-metrics-header')]]</h3>
           this.fire(metricsSignal);
         }
       },
+      _toggleKeysPortButtonsVisible: function(visible) {
```
fortuna: I think "toggle" taking the state is an uncommon pattern. Maybe use `setKeysPortButtonsVisible` instead?

src/server_manager/ui_components/outline-server-settings.html
```diff
@@ -208,6 +245,77 @@ <h3>[[localize('settings-metrics-header')]]</h3>
...
+        this.$.keysPortSaveButton.hidden = !visible;
+        this.$.keysPortCancelButton.hidden = !visible;
+      },
+      _toggleKeysPortButtonsEnabled: function(enabled) {
```
fortuna: Same

src/server_manager/web_app/app.ts
```diff
@@ -757,6 +772,7 @@ export class App {
     view.serverHostname = selectedServer.getHostname();
     view.serverManagementApiUrl = selectedServer.getManagementApiUrl();
     view.serverPortForNewAccessKeys = selectedServer.getPortForNewAccessKeys();
+    view.isAccessKeyPortEditable = await isAccessKeyPortEditable(selectedServer);
```
fortuna: Won't this block the UI? Can we do this where we get the server information instead and make the showServer a sync method again?

src/server_manager/ui_components/outline-server-settings.html
JonathanDCohen: Done.  Also changed for the save and cancel buttons

src/server_manager/ui_components/outline-server-settings.html
JonathanDCohen: Done

src/server_manager/ui_components/outline-server-settings.html
JonathanDCohen: Done

src/server_manager/web_app/app.ts
JonathanDCohen: Done.  We now use the version field in `shadowbox/package.json` to report shadowbox version from the management API.

JonathanDCohen:
Responded, PTAL

src/server_manager/ui_components/outline-server-settings.html
```diff
@@ -208,6 +245,77 @@ <h3>[[localize('settings-metrics-header')]]</h3>
           this.fire(metricsSignal);
         }
       },
+      _setKeysPortButtonsVisible: function(visible) {
```
fortuna: When we create a UI component, we need to think about it's public interface: how do we interact with it.

There are 3 states for an editable input:
- Saved
- Editing (with or without error)
- Saving

There are a few operations:
- Edit (Saved -> Editing, on widget tap)
- Save (Editing -> Saving, on save tap, disables widget and fires event to app)
- CancelEdit (Editing -> Saved, on cancel tap)
- EditInvalidInput (Saving|Editing -> Editing, shows error, called by app)
- Commit (Saving|Editing -> Saved, updates view, called by app)

Let's make sure that's represented in code and create a public method for each of those operations. It will make it easier to follow the logic and allow us to move more code to the Typescript code. Furthermore, it allows for easier testing without having to have do magic with the DOM tree.

It may be worth mentioning the interface above as a comment.

src/server_manager/ui_components/outline-server-settings.html
```diff
@@ -208,6 +245,77 @@ <h3>[[localize('settings-metrics-header')]]</h3>
...
+      _handleKeysPortTap: function() {
+        if (!this.isAccessKeyPortEditable) { 
+          return;
+        }
+        this._setKeysPortButtonsVisible(true);
+        const invalid = this.$.keysPortInput.invalid;
+        this.$.keysPortSaveButton.disabled = invalid;
+        this.$.keysPortCancelButton.disabled = false;
+      },
+      _handleKeysPortServerSuccess() {
```
fortuna: Rename to `commitKeysPort`

src/server_manager/ui_components/outline-server-settings.html
```diff
@@ -208,6 +245,77 @@ <h3>[[localize('settings-metrics-header')]]</h3>
...
+        this.serverPortForNewAccessKeys = Number(this.$.keysPortInput.value);
+        this._setKeysPortButtonsVisible(false);
+        this.$.keysPortInput.readonly = false;
+      },
+      _handleKeysPortServerFail(message) {
```
fortuna: Rename to `editInvalidKeyPort(message)`

src/server_manager/ui_components/outline-server-settings.html
```diff
@@ -245,72 +247,67 @@ <h3>[[localize('settings-metrics-header')]]</h3>
...
+        const badInputMessage = this.localize('error-keys-port-bad-input');
         if (event.key === 'Escape') {
-          this._handleKeysPortCancel();
+          this.cancelKeysPortEdit();
           input.blur();
           return; 
-        } else if (event.key === 'Enter' && !input.invalid) {
-          this._handleKeysPortSave();
+        } else if (event.key === 'Enter') {
+          if (input.invalid && input.errorMessage === badInputMessage) {
```
fortuna: Why do you need to check the error message? Won't the input.invalid give you the information you need?

JonathanDCohen: My mistake, fixed the issue

PR #521, opened 2019-10-22
Update electron for server_manager

JonathanDCohen:
Update electron to v7.0.0 to allow us to use [Spectron](https://github.com/electron-userland/spectron) with Typescript for unit testing UI components -- the [typings for Spectron](https://github.com/electron-userland/spectron/pull/181/commits) were added in Spectron v3.6.3 (`git name-rev --tags --name-only ff92089c298546fe489c852120d72f5b12bf4496` in a Spectron clone), which requires at least Electron 5.0.0.

Changes required for the update:
* `MenuContructorOptions.submenu` now must be a `Menu`, not a `Menu | MenuConstructorOptions[]`

* `MenuConstructorOptions.role` is now a string literal type, and the casing of some options changed.

* `app.makeSingleInstance` was removed for `app.requestSingleInstanceLock()` and the `second-instance` event.

* On Mac, the App Menu "Outline Manager" needed to be given a role

* `electron-builder` and `electron-updater` were also upgraded

* `win.setMaximizable` is now deprecated, so use the `maximizable` option instead.  This was tested and verified to work now on a 2014 MacBookPro running MacOs 10.14.6 (Mojave).

Tested:
* `yarn do shadowbox/integration_test/run`
* `yarn do server_manager/test`


alalamav:
Is this going to be part of the v1.3.0 release? You may want to look into why the unit tests are failing.

JonathanDCohen:
Tests are passing, PTAL

PR #530, opened 2019-11-01
Enable localization

alalamav:
* Imports translated messages for supported languages.
* Enables language detection and automatic localization.
* Ahmaric (am) is not enabled, as it is missing some translations.


alalamav:
Note: do not merge until we get the latest round of AM and KM translations.

Issue #48, opened 2018-03-25
Set Watchtower to selectively watch shadowbox


evanjarrett:
I'll admit I should have read the install script a little more carefully, but while running the install, watchtower tried updating all my other containers, which ended up breaking my wordpress container :(

I would like to suggest only setting watchtower to selectively watch shadowbox.

Reading the watchtower documentation, this can be done by passing in `--label-enable` as a param on watchtower and adding `--label=com.centurylinklabs.watchtower.enable=true` on shadowbox.

More information can be found on the [watchtower repo](https://github.com/v2tec/watchtower#selectively-watching-containers)

trevj:
@00firestar00 Many thanks for the report. I'm happy you got up and running with Outline. Regarding Watchtower settings, we're going to take that into consideration.

alxndrkalinin:
I think I faced the same issue. After reinstalling Outline and eventually restarting the server, my another container got updated and cleaned up :slightly_frowning_face: Would be great if Outline's watchtower didn't mess with existing containers.

AlekSi:
See #1255

PR #540, opened 2019-12-02
Set watchtower to update/watch only shadowbox container

tanelpuhu:
(or others with `com.centurylinklabs.watchtower.enable=true` label). Might fix issue #48. 

Same as in #48 happened to me - after installation many running containers got *upgraded* automatically but thankfully nothing broke. After this change, installed server again and seems to be working just fine, non of the other images have been pulled automatically and containers *upgraded*.


src/server_manager/install_scripts/install_server.sh
```diff
@@ -260,6 +260,7 @@ function start_shadowbox() {
   # rather than pass in the environment.
   declare -a docker_shadowbox_flags=(
     --name shadowbox --restart=always --net=host
+    --label=com.centurylinklabs.watchtower.enable=true
```
JonathanDCohen: A disadvantage of this approach is that existing shadowbox containers won't get the label unless there's a manual re-install.  

Is it possible we could add the label directly to the Dockerfile?  Then when we next  push shadowbox all the existing containers will get this label automatically.

tanelpuhu: Indeed...  if stable image is created from `src/shadowbox/docker/Dockerfile`, perhaps something similar might help:
```
diff --git src/shadowbox/docker/Dockerfile src/shadowbox/docker/Dockerfile
index f5c1ec6..9601f42 100644
--- src/shadowbox/docker/Dockerfile
+++ src/shadowbox/docker/Dockerfile
@@ -24,2 +24,5 @@ LABEL shadowbox.outline-ss-server_version="${SS_VERSION}"

+# Set watchtower to automatically fetch docker image updates
+LABEL com.centurylinklabs.watchtower.enable=true
+
 ARG GITHUB_RELEASE
```
And label here would also be used for custom `SB_IMAGE` but not sure if that is desired behaviour 🤔 maybe it'll be local in most cases...

tanelpuhu: Moved the label to Dockerfile

JonathanDCohen:
LGTM, I'll let you hit the merge button.  Thanks for the contribution!

src/shadowbox/docker/Dockerfile
```diff
@@ -25,6 +25,9 @@ LABEL shadowbox.outline-ss-server_version="${SS_VERSION}"
 ARG GITHUB_RELEASE
 LABEL shadowbox.github.release="${GITHUB_RELEASE}"
 
+# Set watchtower to automatically fetch docker image updates
+LABEL com.centurylinklabs.watchtower.enable=true
```
JonathanDCohen: On discussion with the team offline, we decided that this is too much of a layering violation. Sorry, could you please move this back to the install script?

tanelpuhu: No worries, moved back to install_server.sh

PR #541, opened 2019-12-02
Create an endpoint to change the hostname for new access keys

JonathanDCohen:


src/shadowbox/server/manager_service.spec.ts
```diff
@@ -12,9 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+
```
alalamav: nit: extra line

src/shadowbox/model/access_key.ts
```diff
@@ -59,6 +59,8 @@ export interface AccessKeyRepository {
   listAccessKeys(): AccessKey[];
   // Changes the port for new access keys.
   setPortForNewAccessKeys(port: number): Promise<void>;
+  // Changes the hostname for new access keys.
+  setHostname(hostname: string): void;
```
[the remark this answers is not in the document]
alalamav: I think it's less ambiguous to use `host` instead of `hostname`.  However, I don't feel very strongly about it, since it is also common to use `hostname` to refer to both domain names and IPs. Agreed that we should do the change in a consistent and backward compatible way.

src/shadowbox/server/api.yml
```diff
@@ -27,7 +27,34 @@ paths:
               examples:
                 '0':
                   value: >-
-                    {"name":"My Server","serverId":"000-000-000-000","metricsEnabled":true,"createdTimestampMs":1536613192052,"portForNewAccessKeys":1234}
+                    {"name":"My Server","serverId":"000-000-000-000","metricsEnabled":true,"createdTimestampMs":1536613192052,"portForNewAccessKeys":1234, "version": "1.0.0"}
+  /server/hostname-for-new-access-keys:
```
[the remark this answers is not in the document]
alalamav: I prefer to be more explict and lean towards `hostname-for-access-keys`.

src/shadowbox/server/manager_service.spec.ts
```diff
@@ -110,10 +112,87 @@ describe('ShadowsocksManagerService', () => {
     });
   });
 
+  describe('changeHostname', () => {
```
alalamav: nit: this could be more descriptive "Changes the hostname for access keys"

src/shadowbox/server/manager_service.spec.ts
```diff
@@ -110,10 +112,87 @@ describe('ShadowsocksManagerService', () => {
...
+    const goodHostnames = [
+      "localhost","example.com","www.example.org","www.exa-mple.tw","123abc.co.uk","93.184.216.34","::0","2606:2800:220:1:248:1893:25c8:1946"
+    ];
+    for (const hostname of goodHostnames) {
+      it(`accepts ${hostname}`, (done) => {
```
alalamav: For consistency with the other tests, can we have two top-level tests: `it('accepts valid hostnames')`, `it('rejects invalid hostnames')` and iterate over the hosts inside each test?

src/shadowbox/server/manager_service.ts
```diff
@@ -147,6 +149,37 @@ export class ShadowsocksManagerService {
     next();
   }
 
+  // Changes the server's hostname.  Hostname must be a valid domain or IP address
+  public changeHostname(req: RequestType, res: ResponseType, next: restify.Next): void {
```
alalamav: nit: for consistency, should this be `setHostnameForAccessKeys`?

src/shadowbox/server/manager_service.ts
```diff
@@ -147,6 +149,37 @@ export class ShadowsocksManagerService {
...
+    logging.debug(`changeHostname request: ${JSON.stringify(req.params)}`);
+
+    const hostname = req.params.hostname;
+    if (typeof hostname === 'undefined') {
```
alalamav: nit: `if (!hostname)` to capture `null`

src/shadowbox/server/manager_service.ts
```diff
@@ -147,6 +149,37 @@ export class ShadowsocksManagerService {
...
+      return next(
+        new restify.InvalidArgumentError(
+          {statusCode: 400},
+          `Hostname ${hostname} isn't a valid hostname or IP address`));
+    }
+
+    this.serverConfig.data().hostname = hostname;
+    this.serverConfig.write();
+    this.accessKeys.setHostname(hostname);
+    res.send(204);
```
alalamav: nit: use `HttpSuccess.NO_CONTENT`

src/shadowbox/server/manager_service.spec.ts
JonathanDCohen: This way has the advantage that you get better error messages as you can tell which hostname test failed.  It helped me multiple times during development, so if anything I'd say we should change the others in a separate PR

src/shadowbox/server/api.yml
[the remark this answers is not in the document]
JonathanDCohen: How does it apply to all keys?  Currently we don't do anything to pre-existing keys as there's no way to do so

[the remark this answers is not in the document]
alalamav: @fortuna, please correct me if I'm wrong. The new hostname will appear on all access keys when generating the invite. Of course, pre-existing keys won't be affected.

src/shadowbox/server/manager_service.spec.ts
alalamav: I find it more difficult to parse where the actual test is happening. The manager service error includes the failing hostname and that should appear in the logs.

src/shadowbox/integration_test/test.sh
```diff
@@ -33,6 +33,7 @@
 set -x
 
 export DOCKER_CONTENT_TRUST=${DOCKER_CONTENT_TRUST:-1}
+export CLOUDSDK_PYTHON=/usr/bin/python2
```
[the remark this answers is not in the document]
JonathanDCohen: It's from an error in the Docker stack

src/shadowbox/server/manager_service.spec.ts
JonathanDCohen: Done

src/shadowbox/server/manager_service.spec.ts
JonathanDCohen: Will resolve once we all agree on whether or not this affects all keys or new keys

src/shadowbox/server/manager_service.ts

src/shadowbox/server/manager_service.ts
JonathanDCohen: We capture that in the type guard.  This way is slightly more precise in that we give an error that an empty string is an invalid hostname instead of missing

src/shadowbox/server/manager_service.ts
JonathanDCohen: Waiting to resolve until we decide which access keys are affected

src/shadowbox/server/manager_service.spec.ts
JonathanDCohen: Done

JonathanDCohen:
ping :)

src/shadowbox/integration_test/test.sh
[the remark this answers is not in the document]
fortuna: Our script is cloud-independent. Why does it need a CLOUD env var?
What's the error you are getting?

src/shadowbox/model/access_key.ts
```diff
@@ -59,6 +59,8 @@ export interface AccessKeyRepository {
   listAccessKeys(): AccessKey[];
   // Changes the port for new access keys.
   setPortForNewAccessKeys(port: number): Promise<void>;
+  // Changes the hostname for new access keys.
```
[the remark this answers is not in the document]
fortuna: I still see "new access keys". Did you push the change?

src/shadowbox/server/api.yml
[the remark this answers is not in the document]
fortuna: @alalamav you're right.

src/shadowbox/model/access_key.ts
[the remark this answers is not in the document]
JonathanDCohen: please check again

src/shadowbox/server/manager_service.ts
```diff
@@ -76,6 +77,7 @@ export function bindService(
     apiServer: restify.Server, apiPrefix: string, service: ShadowsocksManagerService) {
   apiServer.put(`${apiPrefix}/name`, service.renameServer.bind(service));
   apiServer.get(`${apiPrefix}/server`, service.getServer.bind(service));
+  apiServer.put(`${apiPrefix}/server/hostname-for-new-access-keys`, service.setHostnameForAccessKeys.bind(service));
```
fortuna: -> `hostname-for-access-keys`

src/shadowbox/integration_test/test.sh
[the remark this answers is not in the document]
fortuna: Please clarify. This is surprising.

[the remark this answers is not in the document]
JonathanDCohen: Removed.  It was an issue with my environment -- docker was finding a python3 binary when it wanted a python 2 binary, apparently during something using google cloud in the implementation.

PR #552, opened 2020-01-08
Return the port for new access keys from GET /server.  

JonathanDCohen:
This was accidentally removed in #537.


JonathanDCohen:
Failure is a known issue with in-development replay attack protection

PR #554, opened 2020-01-09
Data limits UI

alalamav:
* Implements the user interface to configure access key data transfer limits on the Outline Manager app. The feature is currently an experiment, which is why we have a new section in the server settings.

Screen Shot 2020-01-09 at 1 31 58 PM

Screen Shot 2020-01-09 at 1 32 51 PM

* When data limits are enabled, the access key usage display becomes relative to the limit. Displays a tooltip on hover. Re-styles the data usage progress bar.

Screen Shot 2020-01-09 at 1 34 20 PM

* Displays a notification to direct the user to the feature for existing servers upon updating the app, or when opening the app for a second time for new servers.

Screen Shot 2020-01-09 at 1 21 51 PM

* Minor improvements: whitespace removal, turns nested Promise code into async/await, re-styles the help bubbles.


src/server_manager/web_app/app.ts
```diff
@@ -38,6 +38,9 @@ interface PolymerEvent extends Event {
 const UNUSED_DIGITALOCEAN_REFERRAL_CODE = '5ddb4219b716';
 
 const CHANGE_KEYS_PORT_VERSION = "1.0.0";
+const DATA_LIMITS_VERSION = '1.1.0';
```
JonathanDCohen: Do you want to include both this and the hostname stuff together on the next Shadowbox release?

src/server_manager/web_app/app.ts
```diff
@@ -751,7 +767,7 @@ export class App {
   }
 
   // Show the server management screen. Assumes the server is healthy.
-  private showServer(selectedServer: server.Server, selectedDisplayServer: DisplayServer): void {
+  private async showServer(selectedServer: server.Server, selectedDisplayServer: DisplayServer) {
```
JonathanDCohen: nit: missing return type

src/server_manager/ui_components/outline-server-settings.html
```diff
@@ -203,13 +284,18 @@ <h3>[[localize('settings-metrics-header')]]</h3>
...
+        dataLimitsAvailabilityDate: {type: String},
         serverCreationDate: {type: String, value: null},
         serverLocation: {type: String, value: null},
         serverMonthlyCost: {type: String, value: null},
         serverMonthlyTransferLimit: {type: String, value: null},
         localize: {type: Function, readonly: true},
         enableKeysPortButtons: {type: Boolean, value: false},
-        showKeysPortButtons: {type: Boolean, value: false}
+        showKeysPortButtons: {type: Boolean, value: false},
+        shouldShowExperiments: {type: Boolean, computed: '_computeShouldShowExperiments(canSetAccessKeyDataLimit)'},
```
JonathanDCohen: Can we just default this to false and set it in showServer where we set all the other properties anyways?

src/server_manager/web_app/app.ts
```diff
@@ -831,37 +855,49 @@ export class App {
...
+      // tslint:disable-next-line:forin
+      for (const accessKeyId in stats.bytesTransferredByUserId) {
+        totalBytes += stats.bytesTransferredByUserId[accessKeyId];
+      }
+      serverView.setServerTransferredData(totalBytes);
 
+      const accessKeyDataLimit = selectedServer.getAccessKeyDataLimit();
+      if (accessKeyDataLimit) {
+        // Make access key data usage relative to the data limit.
+        totalBytes = accessKeyDataLimit.bytes;
```
JonathanDCohen: Is there a reason we're re-using this variable?  Maybe something like `const dataLimit = accessKeyDataLimit && accessKeyDataLimit.bytes`

src/server_manager/ui_components/outline-server-settings.html
```diff
@@ -311,6 +397,39 @@ <h3>[[localize('settings-metrics-header')]]</h3>
...
+          this.$.accessKeyDataLimitInput.blur();
+        }
+      },
+      _requestSetAccessKeyDataLimit: function() {
+        if (this.$.accessKeyDataLimitInput.invalid) {
+          return;
+        }
+        const value = Number(this.$.accessKeyDataLimitInput.value);
+        const unit = this.$.accessKeyDataLimitUnits.selected;
+        this.fire('SetAccessKeyDataLimitRequested', {limit: {value, unit}});
```
JonathanDCohen: If we pass `serverSettings: this` in these events then we don't have to call appRoot.getServerView in app.ts

src/server_manager/web_app/app.ts
```diff
@@ -920,24 +956,107 @@ export class App {
...
+
+  private dataLimitToDisplayDataAmount(limit: server.DataLimit): DisplayDataAmount|null {
+    if (!limit) {
+      return null;
+    }
+    const bytes = limit.bytes;
+    if (bytes >= 10 ** 9) {
+      return {value: Math.round(bytes / (10 ** 9)), unit: 'GB'};
+    }
+    return {value: Math.ceil(bytes / (10 ** 6)), unit: 'MB'};
```
JonathanDCohen: Do we want to always round down here?  Otherwise users could get cut off when it looks on the UI like they still have quota left.

src/server_manager/web_app/app.ts
```diff
@@ -920,24 +956,107 @@ export class App {
...
+    }
+  }
+
+  private dataLimitToDisplayDataAmount(limit: server.DataLimit): DisplayDataAmount|null {
+    if (!limit) {
+      return null;
+    }
+    const bytes = limit.bytes;
+    if (bytes >= 10 ** 9) {
+      return {value: Math.round(bytes / (10 ** 9)), unit: 'GB'};
```
JonathanDCohen: Also worth noting that due to the rounding that these two conversion functions don't round-trip.  Maybe it's possible instead to do a string conversion and truncate to a maximum length?

src/server_manager/web_app/app.ts
```diff
@@ -920,24 +956,107 @@ export class App {
...
+    try {
+      // Assume non-managed servers have a data transfer capacity of 1TB.
+      let serverTransferCapacity: server.DataAmount = {terabytes: 1};
+      if (isManagedServer(server)) {
+        serverTransferCapacity = server.getHost().getMonthlyOutboundTransferLimit();
+      }
+      if (!accessKeys) {
+        accessKeys = await server.listAccessKeys();
+      }
+      let dataLimitBytes = serverTransferCapacity.terabytes * (10 ** 12) / accessKeys.length;
```
JonathanDCohen: Is it possible for a server to have 0 access keys and error out here? Maybe we want to do (accessKeys.length || 1)

src/server_manager/web_app/app.ts
```diff
@@ -920,24 +956,107 @@ export class App {
...
+    try {
+      await this.selectedServer.setAccessKeyDataLimit(limit);
+      this.appRoot.showNotification(this.appRoot.localize('saved'));
+      serverView.accessKeyDataLimit = this.dataLimitToDisplayDataAmount(limit);
+      this.refreshTransferStats(this.selectedServer, serverView);
+    } catch (error) {
+      console.error(`Failed to set access key data limit: ${error}`);
+      this.appRoot.showError(this.appRoot.localize('error-set-data-limit'));
+      serverView.accessKeyDataLimit = this.dataLimitToDisplayDataAmount(
+          previousLimit || await this.computeDefaultAccessKeyDataLimit(this.selectedServer));
```
JonathanDCohen: IMO it's cleaner if we call `this.computeDefaultAccessKeyDataLimit(this.selectedServer, await this.selectedServer.listAccessKeys())` here and make the second parameter required.  Or alternatively get rid of the second parameter completely and call `listAccessKeys()` twice above

src/server_manager/model/server.ts
```diff
@@ -37,6 +37,15 @@ export interface Server {
   // Removes the access key given by id.
   removeAccessKey(accessKeyId: AccessKeyId): Promise<void>;
 
+  // Sets a data transfer limit over a 30 day rolling window for all access keys.
+  setAccessKeyDataLimit(limit: DataLimit): Promise<void>;
+
+  // Returns the access key data transfer limit, or undefined if it has not been set.
+  getAccessKeyDataLimit(): DataLimit|undefined;
```
fortuna: We should return `|null` instead of `|undefined`, it's more explicit.

We should avoid `undefined` whenever possible. `undefined` is usually an unassigned variable, a return without a value or a missing field in a structure. Things that a usual compiler would complain about. It's an unintentional value.

`null` is an explicit intention to represent the absence of something. You have the guarantee that someone wrote an assignment or a return statement with null. It's well defined.

For those reasons, I believe it's better to use null. However.... I see that we are using `undefined` in other functions, so we may want to keep for consistency, or change all of them. Up to you.

src/server_manager/web_app/app.ts
```diff
@@ -38,6 +38,9 @@ interface PolymerEvent extends Event {
 const UNUSED_DIGITALOCEAN_REFERRAL_CODE = '5ddb4219b716';
 
 const CHANGE_KEYS_PORT_VERSION = "1.0.0";
+const DATA_LIMITS_VERSION = '1.1.0';
+const DATA_LIMITS_AVAILABILITY_DATE = new Date('2020-04-01');
```
fortuna: What is this? Could you add a comment?

src/server_manager/ui_components/outline-server-view.html
```diff
@@ -544,6 +562,10 @@ <h3>[[localize('server-help-access-key-title')]]</h3>
         isAccessKeyPortEditable: {type: Boolean, value: false},
         serverCreationDate: String,
         serverLocation: String,
+        accessKeyDataLimit: {type: Object, value: undefined},
+        isAccessKeyDataLimitEnabled: {type: Boolean},
+        canSetAccessKeyDataLimit: {type: Boolean, value: false},
```
fortuna: Could you add a comment explaining this property and `dataLimitsAvailabilityDate`?

src/server_manager/ui_components/outline-server-view.html
```diff
@@ -758,31 +786,36 @@ <h3>[[localize('server-help-access-key-title')]]</h3>
...
       // they will appear in the top left of the view.
       initHelpBubbles: function() {
-        this.showHelpBubbleOnce('addAccessKeyHelpBubble', 'addAccessKeyRow', 'down', 'left');
+        if (!this.showHelpBubbleOnce('addAccessKeyHelpBubble', 'addAccessKeyRow', 'down', 'left') &&
+            this.canSetAccessKeyDataLimit) {
+          // Show the data limits help bubble if the access key one has been already shown.
+          this.showDataLimitsHelpBubble();
+        }
       },
       showHelpBubbleOnce: function(helpBubbleId, positionTargetId,
```
fortuna: UI components should be dumb and shouldn't have access to storage. Could you move the "once" logic to app.ts instead?

And the `showDataLimitsHelpBubble()` should always show the bubble.

src/server_manager/web_app/app.ts
```diff
@@ -785,18 +806,21 @@ export class App {
...
+    try {
+      const serverAccessKeys = await selectedServer.listAccessKeys();
+      view.accessKeyRows = serverAccessKeys.map(this.convertToUiAccessKey.bind(this));
+      if (!view.accessKeyDataLimit) {
+        view.accessKeyDataLimit = this.dataLimitToDisplayDataAmount(
+            await this.computeDefaultAccessKeyDataLimit(selectedServer, serverAccessKeys));
+      }
+      // Initialize help bubbles once the page has rendered.
+      setTimeout(() => {
+        view.initHelpBubbles();
```
fortuna: The "once" logic should be here, not in the UI component.

src/server_manager/web_app/app.ts
```diff
@@ -920,24 +956,107 @@ export class App {
...
+      this.appRoot.showNotification(this.appRoot.localize('saved'));
+      this.refreshTransferStats(this.selectedServer, serverView);
+    } catch (error) {
+      console.error(`Failed to remove access key data limit: ${error}`);
+      this.appRoot.showError(this.appRoot.localize('error-remove-data-limit'));
+      serverView.isAccessKeyDataLimitEnabled = true;
+    }
+  }
+
+  private dataLimitToDisplayDataAmount(limit: server.DataLimit): DisplayDataAmount|null {
```
fortuna: Move this out of the class, it doesn't depend on `this`. The class is already big enough.

src/server_manager/web_app/app.ts
```diff
@@ -1082,4 +1201,9 @@ export class App {
       this.showCreateServer();
     });
   }
+
+  private localizeDate(date: Date) {
```
fortuna: Move this outside the class and take a language and a Date

src/server_manager/model/server.ts
JonathanDCohen: I had similar thoughts when I first looked at this code but `undefined` does somewhat make sense if the server isn't using data limits.

My actual preference would be to explicitly use an Optional<T> type, something like`type Optional<T> = T|undefined` which would play well with [TS 3.7's null coalescing and optional chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining), and which clearly states the semantics of the returned thing.

src/server_manager/ui_components/outline-server-view.html
alalamav: Done.

src/server_manager/web_app/app.ts
alalamav: Done.

src/server_manager/web_app/app.ts
alalamav: Done in 689d8b9.

src/server_manager/ui_components/outline-server-view.html
alalamav: Done in 689d8b9.

src/server_manager/model/server.ts
alalamav: I have no strong opinions about this, but I do think `undefined` is semantically correct in this case; the server has not defined a value for the data limit and I can see a scenario where null takes a different meaning.

src/server_manager/web_app/app.ts
alalamav: Sure!

src/server_manager/ui_components/cloud-install-styles.html
```diff
@@ -81,7 +81,13 @@
       }
 
       paper-dropdown-menu {
-        width: 100%;
```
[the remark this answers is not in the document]
JonathanDCohen: It uses paper-input, but yeah we're already going to have some merge conflict fun

src/server_manager/web_app/app.ts
JonathanDCohen: I mean as in having the ability to display "1.5 GB" seems reasonable, but we'd have to deal with repeating fractions

src/server_manager/ui_components/outline-server-settings.html
alalamav: We currently don't set any `outline-server-settings` properties in `showServer`. We use the bound properties from `outline-server-view`. We spoke offline about removing the duplication in a separate refactor.

I think computing the property from existing values removes the burden from setting an  additional property in `showServer`.

src/server_manager/ui_components/outline-server-view.html
alalamav: I had added it to `outline-server-settings`, which also declares them. Added here now.

src/server_manager/messages/en.json
```diff
@@ -25,6 +25,11 @@
   "confirmation-server-destroy-title": "Destroy Server?",
   "confirmation-server-remove": "This action removes your server from the Outline Manager, but does not block proxy access to users. You will still need to manually delete the Outline server from your host machine.",
   "confirmation-server-remove-title": "Remove Server?",
+  "data-limits": "Data limits",
+  "data-limits-description": "Set a 30 day trailing data transfer limit for access keys on this server. Available until {date}.",
```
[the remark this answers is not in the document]
alalamav: Done.

src/server_manager/ui_components/outline-server-settings.html
```diff
@@ -176,13 +226,44 @@ <h3>[[localize('settings-server-info')]]</h3>
...
+            </div>
+        </div>
         <!-- Metrics controls -->
         <div class="setting card-section">
           <iron-icon class="setting-icon" icon="editor:insert-chart"></iron-icon>
           <div>
-            <div id="metricsHeader">
+            <div class="space-between">
               <h3>[[localize('settings-metrics-header')]]</h3>
-              <paper-checkbox id="metricsCheckbox" checked="{{metricsEnabled}}"></paper-checkbox>
```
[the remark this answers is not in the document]
fortuna: I've argued this before. The switch is a mobile-only pattern. It's not a desktop pattern.
See https://material.io/components/selection-controls/#usage

> Use switches to:
> - Toggle a single option on or off, on mobile and tablet
> - Immediately activate or deactivate something

> Use checkboxes to:
> - Select one or multiple items from a list
> - Present a list containing sub-selections
> - Turn an option on or off in desktop environment

Here is an example on macOS:
image

Besides, I thought we were going to use an [exposed dropdown menu](https://material.io/components/menus/#usage)?

thelonious89:
UI and initial feature looks great and I look forward to testing~

src/server_manager/ui_components/outline-server-settings.html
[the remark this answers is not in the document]
fortuna: As per offline discussion, we'll proceed with the switch, even though it's in conflict with the spec.

src/server_manager/ui_components/outline-server-settings.html
```diff
@@ -176,13 +226,44 @@ <h3>[[localize('settings-server-info')]]</h3>
             <paper-input readonly value="[[serverId]]" label="[[localize('settings-server-id')]]" hidden$="[[!serverId]]" always-float-label maxlength="100"></paper-input>
           </div>
         </div>
+        <!-- Experiments -->
+        <div id="experiments" class="setting card-section" hidden$="[[!shouldShowExperiments]]">
+            <iron-icon class="setting-icon" icon="icons:build"></iron-icon>
+            <div>
+              <h3>[[localize('experiments')]]</h3>
+              <!-- TODO: link to feedback form -->
+              <p inner-h-t-m-l="[[localize('experiments-description', 'openLink', '<a href=#>', 'closeLink', '</a>')]]"></p>
```
fortuna: inner HTML opens the possibility of vulnerabilities.
It's a good practice to label the messages that contain HTML with a -html suffix to clarify things and avoid mistakes. I have the impression we had established that practice in the client. Can we have the same practice here?

Otherwise, how are you going to prevent the use of text message as HTML or HTML as text messages?

src/server_manager/ui_components/outline-server-settings.html
```diff
@@ -203,13 +284,18 @@ <h3>[[localize('settings-metrics-header')]]</h3>
         serverManagementApiUrl: {type: String, value: null},
         serverPortForNewAccessKeys: {type: Number, value: null},
         isAccessKeyPortEditable: {type: Boolean, value: false},
+        isAccessKeyDataLimitEnabled: {type: Boolean, notify: true},
+        accessKeyDataLimit: {type: Object, value: null}, // type: app.DisplayDataAmount
+        canSetAccessKeyDataLimit: {type: Boolean, value: false}, // Whether the server supports data limits.
```
fortuna: Rename to `supportsAccessKeyDataLimit", it's clearer.

alalamav: Done.

src/server_manager/ui_components/outline-server-settings.html
alalamav: We need use `innerHTML` because of quotation issues when calling `localize`. We do not prefix HTML [messages in the client](https://github.com/Jigsaw-Code/outline-client/blob/master/resources/master_messages.json). We can start doing so, but affected message IDs would have to be re-translated. If this is somehting we want to do anyway, I'd recommend creating an issue and doing it in a separate PR.

fortuna: I filed [b/148406157](http://b/148406157)

alalamav:
@JonathanDCohen, I merged your changes (b1e295b). Tested that the hostname and data limits features are working as intended. Fixed a small UI issue where non-editable hostnames were getting cropped (b1e295b):

Screen Shot 2020-01-30 at 7 16 39 PM

5206d2f and 33eedf1 were UI/UX fixes that resulted from a bug bash with the design team. 

Please take a quick look so I can merge.

cjhenck:
The strings LGTM.

ercxar:
It looks awesome. Thank you.
I would also like to be able to set a different allotment for each user instead of a given limit for all users.

PR #562, opened 2020-01-22
Fix formatting in shadowbox config

### LINKED ISSUE #538: Waiting for Outline server to be healthy
Getting stuck at 
Waiting for Outline server to be healthy
docker logs shadowbox gives

standard_init_linux.go:211: exec user process caused "exec format error"

repeatedely, i am running this in Raspberry Pi 4,  and i am assuming this has to with that  ie; it being ARM pretty strange even running from docker will create this issue, my docker is running fine otherwise.

alalamav:
* The shadowbox server config JSON file was not formatted correctly due incorrectly appending the hostname to the `config` array, which caused `$IFS` in `join` not to take effect.
* Fixes #538.

### DIFF
--- a/src/server_manager/install_scripts/install_server.sh
+++ b/src/server_manager/install_scripts/install_server.sh
@@ -251,7 +251,8 @@ function write_config() {
   if [[ $FLAGS_KEYS_PORT != 0 ]]; then
     config+=("\"portForNewAccessKeys\": $FLAGS_KEYS_PORT")
   fi
-  config+=("\"hostname\": \"$PUBLIC_HOSTNAME\"")
+  # printf is needed to escape the hostname.
+  config+=("$(printf '"hostname": "%q"' ${PUBLIC_HOSTNAME})")
   echo "{"$(join , "${config[@]}")"}" > $STATE_DIR/shadowbox_server_config.json
 }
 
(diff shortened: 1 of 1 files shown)


src/server_manager/install_scripts/install_server.sh
```diff
@@ -249,9 +249,9 @@ function join() {
 function write_config() {
   declare -a config=()
   if [[ $FLAGS_KEYS_PORT != 0 ]]; then
-    config+=("\"portForNewAccessKeys\":$FLAGS_KEYS_PORT")
+    config+=("\"portForNewAccessKeys\": $FLAGS_KEYS_PORT")
   fi
-  config+=$(printf '"hostname": "%q"' ${PUBLIC_HOSTNAME})
```
fortuna: Please revert to using `printf`. That's needed in order to safely escape $PUBLIC_HOSTNAME.

JonathanDCohen: Maybe worth putting a comment about it to since I made the same mistake

alalamav: Reverted to using `printf` with the correct syntax for appending to an array. The missing leading parenthesis added the hostname to `config[0]`, preventing `join`from using the provided separator. Also added a comment.

quocleitxyz: image
Hi all, 
I also have the same error, can't fix it, can you help me

PR #566, opened 2020-01-31
Fix shadowbox/server/build on Cygwin

JonathanDCohen:
rsync treats paths with colons as remote paths, so running `yarn do shadowbox/server/build` on cygwin can result in an error "source and destination must not both be remote"


alalamav:
Does the server run on Windows? Since Windows is not officially supported, I wonder if we should be developing on it.

PR #573, opened 2020-02-11
Add data limits feedback survey link

alalamav:
Directs the user to a data limits survey Google form

FYI, @cjhenck


alalamav:
Friendly ping

JonathanDCohen:
Is there a doc where we figured the questions on the form?  Or could you send an edit link for the form so we can see what it does?

PR #578, opened 2020-02-14
Import localized messages

alalamav:
* Imports translated messages for the data limits and configurable hostname features.
* Backfills the Amharic and Khmer translations with English strings until we resolve an internal translation pipeline issue for these languages.

cc @cjhenck

PR #582, opened 2020-02-20
Restore master branch

alalamav:
* Restores #574, #578, #579 after master was force-pushed at dbd1bb5.
* Reverts dbd1bb5.


JonathanDCohen:
I caused this but thought I had fixed it, sorry about that.  I was trying to fix local branch issues and ended up making the problem worse.

PR #583, opened 2020-02-20
Use Prettier to format HTML and CSS files

JonathanDCohen:
Currently @alalamav uses Sublime's built-in formatter while @JonathanDCohen uses VS Code, causing formatting churn between commits.  This standardizes us on Prettier and adds formatting to our pre-commit hooks.

pretty-quick has no option to only format changed lines, so this PR includes a format of the ui_components/ directory in order to keep unrelated formatting changes from polluting diffs in upcoming commits.

Code contributors can use the instructions at https://prettier.io/docs/en/editors.html if they'd like to integrate prettier with their editor of choice.

Formatting from prettier is in be443b1 and errors which were found by prettier are in b3e51ae


alalamav:
Could you please test that RTL mirroring is not affected by this change? Also, is there any way we can exclude JS (in HTML files) from being formatted - I see some inconsistencies with the current TS/JS style (i.e. double quotes).

JonathanDCohen:
@alalamav double checked in Arabic and RTL looks good to me.

PR #592, opened 2020-02-28
Only lint HTML files with prettier

alalamav:
We don't want prettier to override tslint for TS files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   ],
   "husky": {
     "hooks": {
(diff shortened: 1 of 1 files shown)


JonathanDCohen:
> pretty-quick will respect your .prettierrc, .prettierignore, and .editorconfig files, so there's no additional setup required. **Configuration files will be found by searching up the file system. .prettierignore files are only found from the repository root and the working directory that the command was executed from.**

So I think the issue here is that `src/server_manager/.prettierignore` needs to be at the repo root instead, and then we won't need to specify a pattern here

PR #597, opened 2020-03-02
Expand buttons when their label is too wide to fit

JonathanDCohen:
Before:
fixed

After:
expanded


JonathanDCohen:
I'm leaving the min width in place so the button never gets too small

PR #598, opened 2020-03-03
[feat]: support build for multi-arch builds

oreoluwa:
@JonathanDCohen closing https://github.com/Jigsaw-Code/outline-server/pull/496 in favor of this.
With `seihon` we can easily publish the image for multiple platforms and architectures using the docker manifest feature, while providing a build template. the image that is ultimately deployed would be runnable on any of the supported platforms. By default `seihon` builds for `arm.v6, arm.v7, arm.v8, amd64, 386` however our base image `node:8.15.0-alpine` doesn't have the arm.v7 manifest included, but the v7 architecture can run v6 build.
Please let me know your thoughts on this.


oreoluwa:
@googlebot I consent.

andela-oakinniranye:
@googlebot I consent.

JonathanDCohen:
`seihon` looks really cool, but on https://docs.docker.com/engine/reference/commandline/manifest/ it warns that docker manifests are still experimental and shouldn't be used in production.

@alalamav @fortuna Maybe we could push experimental or unstable manifest lists until `docker manifest` stabilizes?  Otherwise we could do this the hard way without `docker manifest`

And I guess also we should hold until we can get full arm support.  I don't know what the distribution is like in the wild of the different versions of arm devices.

richardzone:
Any update on this?

PR #607, opened 2020-03-30
Display server Shadowbox version

mpmcroy:


alalamav:
Nice change! Let's replace "Shadowbox" with "server" (i.e. "Server version"). The former is an internal name that I don't think we expose to the user.

mpmcroy:
Thanks @alalamav, I updated the message.

PR #610, opened 2020-03-31
Clean up API documentation

JonathanDCohen:


src/shadowbox/server/api.yml
```diff
@@ -192,10 +192,10 @@ paths:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/DataLimit"
+              $ref: "#/components/schemas/AccesKeyName"
```
alalamav: Since this schema is not reflected in the model, it may be more accurate to use an 'untyped' object, i.e.

``` yaml
application/json:
  schema:
    type: object
    properties:
      name:
        type: string
```

JonathanDCohen: Done

PR #628, opened 2020-04-28
Fix sentry

fortuna:
I'm sorry for the breakage!


fortuna:
Submitting TBR, since master is broken. Feel free to add comments

PR #644, opened 2020-05-09
Upgrade restify to v8.5.1

alalamav:
- Updates restify and @types/restify to v8.
- Replaces core restify CORS with restify-cors-middleware, as it has been removed from restify. 
  - The options we pass are the same as the previous defaults.
  - This generates a warning: `restify-cors-middleware@1.1.1" has incorrect peer dependency "restify@2.6.x - 7.x.x"`; however, I haven't seen related runtime errors.
  - I'm not sure I fully understand why we need CORS, since we configure it with any origin (`"*"`).

### DIFF
--- a/src/shadowbox/package.json
+++ b/src/shadowbox/package.json
@@ -16,13 +16,16 @@
     "prom-client": "^11.1.3",
     "randomstring": "^1.1.5",
     "request-lite": "^2.40.1",
-    "restify": "^4.3.0",
+    "restify": "^8.5.1",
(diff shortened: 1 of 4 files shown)


alalamav:
Got it. The CORS configuration was missing a step. I've fixed it in ff303da and tested that it works with the manager app.

alalamav:
The integration test was failing because the restify body parser was not populating the request object with parameters. Fixed in bf8ff94.

PR #665, opened 2020-05-28
Replace request-lite with follow-redirects

alalamav:
* Removes request-lite dependency, as request (its parent module) is now deprecated. Adds follow-redirects as a replacement.
* Fixes a failure with reporting metrics, observed on shadowbox v1.2.1.
* Fixes an unhandled promise rejection in reporting feature metrics.


alalamav:
Sounds good. The metrics server redirects with 307 to upgrade HTTP to HTTPs, so follow-redirects does work for our use case. I didn't test with 302s.

PR #686, opened 2020-06-09
Separate server management from server provisioning

mpmcroy:
**Refactor to prep for adding more "magic" cloud providers**

This is a first pass to separate the cloud provisioning logic from server management logic into their own (logical) _apps_. These are two disparate flows that can likely be isolated to make the manager architecture cleaner.

I'm thinking that we should draw from domain driven design and define the [bounded context](https://martinfowler.com/bliki/BoundedContext.html) for each app. This would give us a vocabulary that makes sense for the separate flows.

**What this might look like:**
Server Management App server with the following concepts:
* Server information (name, id, version)
* Stats (data transferred)
* Management URL
* Access keys
* Experiments (like data limits)

Server Provisioning App:
* Managed vs Manual servers
* Shadowbox server
* DigitalOcean provisioning logic (which could later be it's own app/module)

I feel like we already implicitly have this, but we could benefit from making it more formal.

This becomes a little tricky (at least for me) when trying to separate the UI logic. We rely on state to keep track of things like _"which server is selected"_ and _"which server is being provisioned"_. So for example, when we fire events like _AddAccessKeyRequested_ or _ServerRenameRequested_, we check which server is currently selected to know which one to update.

My concern with this pattern is that we use this state in both server management and server provisioning flows, so I can't easily tell if it's safe to update that value. In fact, we have a few "sanity check" conditional statements just to make sure we're in the proper state.

I took a stab at factoring out the server management UI by moving it into a separate class and refactoring some of the method level logic. I think what I have below gives a rough idea of what this separation could look like with a few caveats:
* I removed two sanity checks because they depended on state that the new server management app no longer has access to.
* Event bindings still not entirely separated so they're still in _app.ts_

### DIFF
--- a/src/server_manager/web_app/ui_components/outline-server-settings.js
+++ b/src/server_manager/web_app/ui_components/outline-server-settings.js
@@ -299,7 +299,10 @@ Polymer({
     }
     // Fire signal if name has changed.
     if (newName !== this.initialName) {
-      this.fire('ServerRenameRequested', {newName});
+      this.fire('ServerRenameRequested', {
+        newName,
+        displayServer: this._makeDisplayServer()
+      });
     }
   },
 
@@ -327,7 +330,9 @@ Polymer({
     if (isDataLimitEnabled) {
       this._requestSetAccessKeyDataLimit();
     } else {
-      this.fire('RemoveAccessKeyDataLimitRequested');
+      this.fire('RemoveAccessKeyDataLimitRequested', {
+        displayServer: this._makeDisplayServer()
+      });
     }
   },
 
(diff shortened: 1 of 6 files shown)


src/server_manager/web_app/ui_components/outline-server-settings.js
```diff
@@ -299,7 +299,7 @@ Polymer({
     }
     // Fire signal if name has changed.
     if (newName !== this.initialName) {
-      this.fire('ServerRenameRequested', {newName});
+      this.fire('ServerRenameRequested', {newName, displayServer: this._makeDisplayServer()});
```
mpmcroy, line 302: To make it more explicit about which server to rename, we would pass along the information in the event.

We're passing the whole _DisplayServer_ object in this PR, but this can be simplified to just the server ID.

PR #693, opened 2020-06-12
Fix unit tests

alalamav:
- Since migrating to webpack the unit tests have not been running. This is due to the fact that we create a single JS bundle, whereas jasmine expects individual JS spec files.
- Fixes shadowbox and the manager unit testing by transpiling individual files via `tsc` (no `webpack`).
- The manager web app tests for `app.spec.ts` are still not running because  `app.ts` dependencies (notably polymer) target the browser and use ES6 imports, which is not compatible with jasmine.


src/server_manager/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
```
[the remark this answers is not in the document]
alalamav: I don't think so - sharing code between workspaces should be orthogonal to the build configuration of each package.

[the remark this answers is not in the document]
fortuna: Also: you can't share code that isn't shareable. If code depends on the broswer environment, you can't run on Node. If code depends on the Node environment, you can't run on the browser. There are polyfills for some things, but that's limited.

src/server_manager/electron_app/test_action.sh
```diff
@@ -14,5 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-do_action server_manager/electron_app/build
+rm -rf $BUILD_DIR/js
```
fortuna, line 17: I think we should delete only `$BUILD_DIR/js/server_manager/electron_app`
An action shouldn't interfere with another. I could imagine a root test action that calls all the other test actions

src/server_manager/tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+  "extends": "../../tsconfig.json",
+  "rootDir": ".",
+  "include": [
+    "**/*.ts"
```
fortuna, line 5: You could actually use webpack, and make each test file an entrypoint, outputting them with the file-loader. This way you wouldn't have to worry about the exclude below.

src/shadowbox/test_action.sh
```diff
@@ -14,5 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-do_action shadowbox/server/build
+rm -rf $BUILD_DIR/js
```
fortuna, line 17: Same as before: only delete what this script creates

src/server_manager/electron_app/test_action.sh
JonathanDCohen: If we want to take this out further, we could `export BUILD_DIR=$(mktemp -d)` in our test actions so we can completely isolate.  Maybe that's overkill?

src/shadowbox/test_action.sh
alalamav: Done.

src/server_manager/tsconfig.json
alalamav: Good suggestion. I have a prototype, but we still have to exclude the web app tests from the webpack build. Happy to follow up on this.

src/server_manager/electron_app/test_action.sh
alalamav: I agree with that, the reason I removed the entire test directory is to prevent jasmine from running previously built tests, since we have a single jasmine config.

I now delete each action test directory after (succesfully) running the tests.

PR #696, opened 2020-06-15
Adds app language selection picker

mpmcroy:
**Adds the ability to change the app language**

Outline Manager currently uses the Chrome language setting (navigator.language) as the language used to display strings. This PR adds the ability to override that default by selecting a preferred language from the side drawer. On subsequent runs, we first check for a preferred language or default to the browser language if none is found.

Below are screenshot of the side drawer language selector:
1) when the side drawer is first opened
1) when the dropdown is focused
1) when the dropdown is expanded
1) in an rtl language

Screen Shot 2020-06-15 at 4 34 56 PM
Screen Shot 2020-06-15 at 4 35 14 PM
Screen Shot 2020-06-15 at 4 35 22 PM
Screen Shot 2020-06-15 at 4 35 36 PM


src/server_manager/web_app/app.ts
```diff
@@ -247,6 +247,10 @@ export class App {
       }
     });
 
+    appRoot.addEventListener('ChangeLanguage', (event: CustomEvent) => {
```
fortuna, line 250: Maybe name the event `SetLanguageRequested`? The language hasn't changes yet.

src/server_manager/web_app/ui_components/app-root.js
```diff
@@ -229,12 +229,49 @@ export class AppRoot extends mixinBehaviors
...
         outline: none;
       }
+      #links-footer {
+        margin-top: 36px;
+      }
+      #language-icon {
+        padding-top: 10px;
+      }
+      #language-dropdown {
+        padding-left: 22px;
```
fortuna, line 239: This is not RTL-friendly. Does the left padding need to be different from the right padding?

src/server_manager/web_app/ui_components/app-root.js
```diff
@@ -229,12 +229,49 @@ export class AppRoot extends mixinBehaviors
...
       }
+      #language-row {
+        display: flex;
+        align-items: center;
+      }
+      .language-item {
+        display: flex;
+        cursor: pointer;
+        font-size: 16px;
+        padding-left: 24px;
```
fortuna, line 265: Same: can the padding be symetric?

src/server_manager/web_app/ui_components/app-root.js
```diff
@@ -229,12 +229,49 @@ export class AppRoot extends mixinBehaviors
...
+        --paper-item-selected: {
+          color: var(--primary-green);
+          font-weight: normal;
+        }
+      }
+      .language-name {
+        text-align: left;
```
fortuna, line 272: Is this needed? Won't the dir property cause this to be set appropriately?

src/server_manager/web_app/ui_components/app-root.js
```diff
@@ -874,6 +933,20 @@ export class AppRoot extends mixinBehaviors
     this.maybeCloseDrawer();
   }
 
+  _sortLanguageNameAscending(a, b) {
```
fortuna, line 936: Sort when you set the language list. It's easier to understand and doesn't rely on Polymer (and maybe one may want to use a different order).

src/server_manager/web_app/ui_components/app-root.js
```diff
@@ -526,17 +580,20 @@ export class AppRoot extends mixinBehaviors
...
+  setLanguage(language, direction) {
+    const messagesUrl = `./messages/${language}.json`;
+    this.loadResources(messagesUrl, language);
+
+    document.documentElement.setAttribute('dir', direction);
+    const alignDir = direction === 'ltr' ? 'left' : 'right';
+    this.$.appDrawer.align = alignDir;
+    this.$.sideBar.align = alignDir;
+
+    window.localStorage.setItem('overrideLanguage', language);
```
fortuna, line 595: Humm... It's better if the UI components don't know about localStorage. It ties the component to an application. It's a hidden global dependency.

Can you move this logic to app.ts?
Same for the document.documentElement attribute, since it's not part of the AppRoot element..

src/server_manager/web_app/ui_components/app-root.js
```diff
@@ -367,15 +404,30 @@ export class AppRoot extends mixinBehaviors
...
             <span on-tap="submitFeedbackTapped">[[localize('nav-feedback')]]</span>
-            <a href="https://s3.amazonaws.com/outline-vpn/index.html#/en/support/">[[localize('nav-help')]]</a>
+            <span on-tap="maybeCloseDrawer"><a href="https://s3.amazonaws.com/outline-vpn/index.html#/en/support/">[[localize('nav-help')]]</a></span>
             <span on-tap="aboutTapped">[[localize('nav-about')]]</span>
-            <div class="legal-links">
-              <a href="https://www.google.com/policies/privacy/">[[localize('nav-privacy')]]</a>
-              <a href="https://s3.amazonaws.com/outline-vpn/static_downloads/Outline-Terms-of-Service.html">[[localize('nav-terms')]]</a>
-              <span on-tap="showLicensesTapped">[[localize('nav-licenses')]]</span>
+            <div id="links-footer">
+              <paper-icon-item id="language-row">
```
fortuna, line 413: It seems like you could factor out this component, in order to not grow app-root further. Maybe use LitElement?

src/server_manager/web_app/ui_components/app-root.js
```diff
@@ -367,15 +404,30 @@ export class AppRoot extends mixinBehaviors
...
+                <iron-icon id="language-icon" icon="language" slot="item-icon"></iron-icon>
+                <paper-dropdown-menu id="language-dropdown" no-label-float="" vertical-align="bottom">
+                  <paper-listbox slot="dropdown-content" selected="{{language}}" attr-for-selected="value" on-selected-changed="_languageChanged">
+                    <template is="dom-repeat" items="{{supportedLanguages}}" as="lang" sort="_sortLanguageNameAscending">
+                      <paper-item class="language-item" value="{{lang.id}}">
+                        <span class="language-name">{{lang.name}}</span>
+                        <iron-icon icon="check" hidden$="{{_shouldHideLanguageCheckmark(language, lang.id)}}"></iron-icon>
```
fortuna, line 420: Is the checkmark working on RTL?

src/server_manager/web_app/ui_components/app-root.js
```diff
@@ -874,6 +933,20 @@ export class AppRoot extends mixinBehaviors
...
+  _sortLanguageNameAscending(a, b) {
+    return a.name > b.name ? 1 : -1;
+  }
+
+  _shouldHideLanguageCheckmark(language, languageCode) {
+    return language !== languageCode;
+  }
+
+  _languageChanged(event) {
+    const code = event.detail.value;
```
fortuna, line 945: This should be firing the 'SetLanguageRequested' event instead. You should only need the language code. The handling should know about the direction. The component doesn't display the direction.

src/server_manager/web_app/app.ts
mpmcroy: Done

src/server_manager/web_app/ui_components/app-root.js
mpmcroy: Done

src/server_manager/web_app/ui_components/app-root.js
mpmcroy: Converted it to Polymer temporarily. Will look into converting to LitElement.

src/server_manager/web_app/ui_components/app-root.js
mpmcroy: Yes!
Screen Shot 2020-06-16 at 1 02 11 PM

src/server_manager/web_app/ui_components/app-root.js
mpmcroy: Done

src/server_manager/web_app/ui_components/app-root.js
mpmcroy: Let me look into it a bit. I didn't do anything special because it looked to be already handled.
Screen Shot 2020-06-16 at 1 09 07 PM

src/server_manager/web_app/ui_components/app-root.js
mpmcroy: Good catch, it's not! Removed.

src/server_manager/web_app/ui_components/app-root.js
mpmcroy: We do this in a few places so I assumed it was safe. It seems to be taken care of by our RTL mixin, if I'm not mistaken.

src/server_manager/web_app/ui_components/app-root.js
mpmcroy: Done

src/server_manager/web_app/ui_components/outline-language-picker.js
```diff
@@ -0,0 +1,94 @@
...
+import '@polymer/paper-listbox/paper-listbox.js';
+import '@polymer/paper-item/paper-item.js';
+import '@polymer/iron-icon/iron-icon.js';
+import '@polymer/iron-icons/iron-icons.js';
+import './cloud-install-styles.js';
+
+import {html, PolymerElement} from '@polymer/polymer';
+import {mixinBehaviors} from '@polymer/polymer/lib/legacy/class.js';
+
+export class OutlineLanguagePicker extends mixinBehaviors
```
fortuna, line 24: Why do you need mixinBehaviors?

src/server_manager/web_app/main.ts
```diff
@@ -103,6 +107,8 @@ document.addEventListener('WebComponentsReady', () => {
   // NOTE: this cast is safe and allows us to leverage Polymer typings since we haven't migrated to
   // Polymer 3, which adds typescript support.
   const appRoot = document.getElementById('appRoot') as unknown as AppRoot;
+
+  appRoot.supportedLanguages = Object.values(SUPPORTED_LANGUAGES);
```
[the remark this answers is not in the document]
mpmcroy: Done

src/server_manager/web_app/ui_components/app-root.js
```diff
@@ -229,12 +230,28 @@ export class AppRoot extends mixinBehaviors
       #appDrawer a:focus {
         outline: none;
       }
+      #links-footer {
+        margin-top: 36px;
+      }
       .legal-links {
-        margin: 36px -6px 0;
+        margin: 0 -6px 0;
```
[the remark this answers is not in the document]
mpmcroy: Done

src/server_manager/web_app/ui_components/outline-language-picker.js
mpmcroy: Not needed. Thanks

mpmcroy:
@fortuna I removed 'zh' from the list because it seems to be identical to 'zh-cn' (Chinese simplified)

src/server_manager/web_app/main.ts
```diff
@@ -103,6 +114,10 @@ document.addEventListener('WebComponentsReady', () => {
   // NOTE: this cast is safe and allows us to leverage Polymer typings since we haven't migrated to
   // Polymer 3, which adds typescript support.
   const appRoot = document.getElementById('appRoot') as unknown as AppRoot;
+
+  const filteredLanguageDefs = Object.values(SUPPORTED_LANGUAGES)
+      .filter((languageDef) => languageDef.id !== 'zh');
```
fortuna, line 119: Let's remove zh from the list. And maybe remove the message file too.
@alalamav could you weigh in?

src/server_manager/web_app/ui_components/outline-language-picker.js
```diff
@@ -0,0 +1,92 @@
...
+      selectedLanguage: {type: String},
+      // An array of {id, name, dir} language objects.
+      languages: {type: Array, readonly: true},
+    };
+  }
+
+  constructor() {
+    super();
+    this.selectedLanguage = '';
+    this.languages = [];
```
fortuna, line 75: Add a @type JSDoc here. I think you can use the TS syntax for the type:
`/** @type {{id: string, name: string, dir: "rtl"|"ltr"}} */

src/server_manager/web_app/main.ts
alalamav: Keeping the more specific variant (zh-CN) sounds good to me. We should also update the translation configuration to prevent additional translations (happy to do it).

mpmcroy: Sounds good. I removed the 'zh' language code and associated messages file.

@alalamav I'll sync with you on the translation configuration so that I can learn how that works.

src/server_manager/web_app/ui_components/outline-language-picker.js
mpmcroy: Done

PR #700, opened 2020-06-17
Use the non-deprecated Buffer factory function

JonathanDCohen:
We don't run into any of the security issues associated with `new Buffer()`, which mostly involve uninitialized memory, but this will hopefully clean up our build output a little bit.

Tested:
  Installed a server


JonathanDCohen:
@fortuna you mean the manual install script?

PR #728, opened 2020-09-15
Seems to Fix a problem of "mkdir permission denied"

Ushio-Cheng:
Try run this script on Debian 10, gives error "mkdir: cannot create directory ‘/opt/outline’: Permission denied".
This seems to fix that issue.
`--mode` just does not seems to appear in man page. Maybe this works on another distro but I think `-m` might be more compatiable.


Ushio-Cheng:
@googlebot I signed it!

PR #730, opened 2020-09-16
Adds DigitalOcean create server "app"

mpmcroy:
Refactors the DigitalOcean server creation logic and UI into a separate LitElement web component (app). Most of this logic was spread across multiple files (_app.ts_, _app_root_.js, _outline_region_picker.ts_, _outline-do-oauth-step.js_, etc). This app is solely for creating Outline servers on DigitalOcean, but similar apps are planned for other cloud providers.

## DigitalOcean Create Server app
1) Account verification (email, billing)
1) Location selection

Screen Shot 2020-10-18 at 12 43 52 PM

Screen Shot 2020-10-18 at 12 42 38 PM

server_create_app

### Interface
`start(digitalOceanServerRepository: DigitaloceanServerRepository): Promise<void>`

_Note: DigitaloceanServerRepository to be replaced with DigitalOceanAccount_

#### App Events
* server-created
  * { server: ManagedServer }
* server-create-cancelled

#### Domain Events
api_connectivity_issue: Event that signals an issue connecting to the DigitalOcean API. This usually means an invalid authentication, CORS, or network issue.

### Misc
* Removes digitalOceanRetry


src/server_manager/web_app/app.ts
```diff
@@ -137,10 +124,37 @@ type DigitalOceanSessionFactory = (accessToken: string) => digitalocean_api.Digi
 type DigitalOceanServerRepositoryFactory = (session: digitalocean_api.DigitalOceanSession) =>
     server.ManagedServerRepository;
 
+export class NotificationManager {
+  // @ts-ignore
+  constructor(private toast) {}
```
mpmcroy, line 129: What is the type of this and how do I find out?

src/server_manager/web_app/ui_components/digitalocean_create_server_app.ts
```diff
@@ -0,0 +1,83 @@
...
+  @property({type: Object}) digitalOceanRepository: server.ManagedServerRepository = null;
+  @property({type: Object}) notificationManager: NotificationManager = null;
+
+  private regionPicker: OutlineRegionPicker;
+
+  // @ts-ignore
+  firstUpdated(_changedProperties) {
+    super.firstUpdated(_changedProperties);
+
+    this.regionPicker = new OutlineRegionPicker();
```
mpmcroy, line 49: I'm adding this programmatically for now because it's not loading properly when added in the render function. I suspect that it is because the localize function is not available at the time.

src/server_manager/web_app/app.ts
fortuna: That's a <paper-toast> element. It's defined in the approot html.

src/server_manager/web_app/app.ts
```diff
@@ -137,10 +124,37 @@ type DigitalOceanSessionFactory = (accessToken: string) => digitalocean_api.Digi
 type DigitalOceanServerRepositoryFactory = (session: digitalocean_api.DigitalOceanSession) =>
     server.ManagedServerRepository;
 
+export class NotificationManager {
```
fortuna, line 127: Can you make this its own component outside approot with the show error and show notifications methods?

src/server_manager/web_app/app.ts
```diff
@@ -727,25 +741,13 @@ export class App {
...
-              console.error(`Failed to get list of available regions: ${e}`);
-              this.appRoot.showError(this.appRoot.localize('error-do-regions'));
-            });
+  private async showCreateServer() {
+    const digitalOceanCreateServer =
+        this.appRoot.getAndShowDigitalOceanCreateServer() as DigitalOceanCreateServer;
+    digitalOceanCreateServer.app = this;
+    digitalOceanCreateServer.digitalOceanRepository = this.digitalOceanRepository;
+    digitalOceanCreateServer.notificationManager = new NotificationManager(this.appRoot);
+    await digitalOceanCreateServer.show();
```
fortuna, line 750: Pass the parameters above to `show()` instead. This way we make sure they are always set on show.

src/server_manager/web_app/digitalocean_app/create_server_app.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  // The region picker initially shows all options as disabled. Options are enabled by this code,
+  // after checking which regions are available.
+  async show() {
+    const regionPicker = this.shadowRoot.querySelector('#regionPicker') as OutlineRegionPicker;
+    regionPicker.reset();
+
+    try {
+      const map =
+          await this.app.digitalOceanRetry(() => this.digitalOceanRepository.getRegionMap());
```
fortuna, line 67: Can you remove the dependency on app , since all you need is `digitalOceanRetry`?

src/server_manager/web_app/app.ts
mpmcroy: Do we have typing information for it? I don't know what type to assign it.

src/server_manager/web_app/digitalocean_app/create_server_app.ts
```diff
@@ -0,0 +1,77 @@
...
+@customElement('digital-ocean-create-server')
+export class DigitalOceanCreateServer extends LitElement {
+  @property({type: Function}) localize: Function;
+  @property({type: Object}) app: App = null;
+  @property({type: Object}) digitalOceanRepository: server.ManagedServerRepository = null;
+  @property({type: Object}) notificationManager: NotificationManager = null;
+
+  render() {
+    return html`
+        <outline-region-picker-step id="regionPicker" .localize=${
```
fortuna, line 45: I'm wondering if we still need the region picker component. Maybe we could have all the logic here to reduce the layering and simplify the logic.

For example, you can probably get rid of the createLocationModel below.

fortuna (changes requested):

src/server_manager/web_app/digitalocean_app/create_server_app.ts
mpmcroy: Absolutely.

I updated the code to pass in the retry function, but I'd like to make that function more generic in the future. 

I moved the digital ocean specific error handling into the DO API, but the retry function also has a UI. We should be able to reuse the same connectivity error dialog for other cloud providers once we make the underlying UI business logic generic.

Screen Shot 2020-09-18 at 9 53 32 AM

src/server_manager/web_app/app.ts
mpmcroy: 👍 

I think the cleanest interface would be something like `DigitalOceanCreateServer#show(account: Account)` (where in this case, the `digitalOceanRepository` acts as the account model until we finish the refactor.

It's probably best to register the `NotificationView` when we construct the `DigitalOceanCreateServer`, so I'll use data binding in appRoot... _as long as I can find the right lifecycle methods to use._

src/server_manager/web_app/digitalocean_app/create_server_app.ts
```diff
@@ -0,0 +1,73 @@
...
+  lon: `${FLAG_IMAGE_DIR}/uk.png`,
+  sfo: `${FLAG_IMAGE_DIR}/us.png`,
+  tor: `${FLAG_IMAGE_DIR}/canada.png`,
+  nyc: `${FLAG_IMAGE_DIR}/us.png`,
+};
+
+@customElement('digital-ocean-create-server')
+export class DigitalOceanCreateServer extends LitElement {
+  @property({type: Function}) localize: Function;
+  @property({type: Object}) notificationView: OutlineNotificationView = null;
```
mpmcroy, line 38: @fortuna: I'm not sure that I like calling this a `view` because it's more than just the UI. I'm leaning towards `OutlineNotificationComponent` instead. Thoughts?

_Naming is hard 😖_

src/server_manager/web_app/digitalocean_app/create_server_app.ts
mpmcroy: Let me take a stab at the GCP create server app and see if we can reuse the `OutlineRegionPicker`. If we can, then it might be good to keep this it cloud agnostic.

Also, I'm leaning towards the adding the create server progress screen to this component. Can you explain again why you think it should be part of the server view?

src/server_manager/web_app/digitalocean_app/create_server_app.ts
fortuna: I'm not a big fan of appending "Component". It's too generic (can be dropped?).
I was looking at the Open WC codelab. They use App for the app, and nothing for the other elements: https://open-wc.org/codelabs/intermediate/lit-html.html?index=/codelabs/#9

Perhaps something like NotificationManager, NotificationService, Notifications?

src/server_manager/web_app/app.ts
```diff
@@ -519,7 +511,8 @@ export class App {
                 this.showIntro();
                 const msg = `Failed to get DigitalOcean account information: ${error}`;
                 console.error(msg);
-                this.appRoot.showError(this.appRoot.localize('error-do-account-info'));
+                this.appRoot.getNotificationView().showError(
```
fortuna, line 514: `this.notificationManager.showError()`?

src/server_manager/web_app/app.ts
```diff
@@ -624,32 +617,26 @@ export class App {
   //       return the UI to its exact prior state. Fortunately, the most likely
   //       time to discover an invalid access token is when the application
   //       starts.
-  private digitalOceanRetry = <T>(f: () => Promise<T>): Promise<T> => {
-    return f().catch((e) => {
-      if (!(e instanceof digitalocean_api.XhrError)) {
-        return Promise.reject(e);
-      }
+  digitalOceanRetry<T>(fn: () => Promise<T>): Promise<T> {
```
fortuna, line 620: What happened to the idea of emitting an event from  the repo instead?

This flow is a bit convoluted.

src/server_manager/web_app/app.ts
```diff
@@ -727,25 +714,11 @@ export class App {
...
-              });
-              regionPicker.locations = locations;
-            },
-            (e) => {
-              console.error(`Failed to get list of available regions: ${e}`);
-              this.appRoot.showError(this.appRoot.localize('error-do-regions'));
-            });
+  private async showCreateServer() {
+    const digitalOceanCreateServer =
+        this.appRoot.getAndShowDigitalOceanCreateServer() as DigitalOceanCreateServer;
```
fortuna, line 719: Do you need the "as"?
Perhaps put the type in the Javascript code? That function should be properly typed.

src/server_manager/web_app/digitalocean_app/create_server_app.ts
```diff
@@ -0,0 +1,75 @@
...
+        <outline-region-picker-step id="regionPicker" 
+            .localize=${this.localize}></outline-region-picker-step>
+    `;
+  }
+
+  // The region picker initially shows all options as disabled. Options are enabled by this code,
+  // after checking which regions are available.
+  async show(
+      digitalOceanRepository: ManagedServerRepository,
+      retryFn: <T>(fn: () => Promise<T>) => Promise<T>) {
```
fortuna, line 51: The idea we discussed before was to not do a retry here, and instead emit an auth error that the app.ts would handle. That would make the flow cleaner and easier to understand. Also more consistent with the rest of the app.

Did you figure out if you can detect auth errors?

src/server_manager/web_app/digitalocean_app/create_server_app.ts
```diff
@@ -0,0 +1,75 @@
...
+  lon: `${FLAG_IMAGE_DIR}/uk.png`,
+  sfo: `${FLAG_IMAGE_DIR}/us.png`,
+  tor: `${FLAG_IMAGE_DIR}/canada.png`,
+  nyc: `${FLAG_IMAGE_DIR}/us.png`,
+};
+
+@customElement('digital-ocean-create-server')
+export class DigitalOceanCreateServer extends LitElement {
+  @property({type: Function}) localize: Function;
+  @property({type: Object}) notificationManager: OutlineNotificationManager = null;
```
fortuna, line 38: You can remove the annotation for this one, since you never set it in HTML.

src/server_manager/web_app/ui_components/app-root.js
```diff
@@ -655,48 +647,20 @@ export class AppRoot extends mixinBehaviors
...
-        text: message,
-        duration: duration,
-        noOverlap: true,
-      });
-    }, 0);
-  }
-
-  closeError() {
-    this.$.toast.close();
+  getNotificationManager() {
```
fortuna, line 650: Add a return type

src/server_manager/web_app/ui_components/app-root.js
```diff
@@ -588,10 +581,9 @@ export class AppRoot extends mixinBehaviors
     return oauthFlow;
   }
 
-  getAndShowRegionPicker() {
-    this.currentPage = 'regionPicker';
-    this.$.regionPicker.reset();
-    return this.$.regionPicker;
+  getAndShowDigitalOceanCreateServer() {
```
fortuna, line 584: Add a return type

src/server_manager/web_app/digitalocean_app/create_server_app.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  render() {
+    return html`
+        <outline-region-picker-step id="regionPicker"
```
fortuna, line 42: Could you handle the click on "Set up Outline" here?

The repo can emit a domain even "ServerCreated(server)", which app.ts should listen to and show the server when that happens.

fortuna (commented):
This looks mostly good. I like the new NotificationManager.

There are 2 things that we talked about before that could make the code clearer:
- Use domain events for the authentication failures and server creation
- Get rid of the retry.

On authentication error the app.ts would trigger a dialog.
On server creation the app.ts would proceed with the flow.

src/server_manager/web_app/ui_components/app-root.js
mpmcroy: Done

src/server_manager/web_app/ui_components/app-root.js
mpmcroy: Done

src/server_manager/cloud/digitalocean_api.ts
```diff
@@ -251,3 +251,15 @@ function makeValidDropletName(name: string): string {
   // Remove all characters outside of A-Z, a-z, 0-9 and '-'.
   return name.replace(/[^A-Za-z0-9\-]/g, '');
 }
+
+export async function digitalOceanRetry<T>(
```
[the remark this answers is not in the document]
mpmcroy: It is with the recursive call, but based on our brainstorming, this will change!

My plan is to throw a domain event from the Account (ManagedServerRepository) and handle or rethrow in the individual apps (DigitalOceanCreateServerApp). Atm, I can't differentiate between a network error and an authorization error, but I will try @alalamav 's suggestion and check out the CORS middleware layer we added for DigitalOcean.

src/server_manager/web_app/digitalocean_app/create_server_app.ts
mpmcroy: I'm not sure why setting the property isn't working. I suspect that it's because the `OutlineNotificationManager` isn't ready when we do the property binding. Have you seen this behavior before? Do you have any ideas how to make this work with property binding?

If not, we can just add a separate `initialize(...)` method (or setter)?

mpmcroy (commented):
Thanks for the review, Vini!

I updated this PR based on our brainstorm. It's almost ready for another pass. I'll let you know when 🙏

src/server_manager/web_app/app.ts
mpmcroy: Thoughts on whether this should be an app event or domain event?

src/server_manager/cloud/digitalocean_api.ts
```diff
@@ -220,6 +220,8 @@ class RestApiSession implements DigitalOceanSession {
           // this.response may be empty.
           const responseObj = (xhr.response ? JSON.parse(xhr.response) : {});
           resolve(responseObj);
+        } else if (xhr.status === 401) {
+          reject(new XhrError());
```
fortuna, line 224: The handling below seems more useful,  since it raises an error with more information.

src/server_manager/cloud/digitalocean_api.ts
```diff
@@ -220,6 +220,8 @@ class RestApiSession implements DigitalOceanSession {
...
         } else {
           // this.response is a JSON object, whose message is an error string.
           const responseJson = JSON.parse(xhr.response);
```
fortuna, line 227: This may throw if the response is not JSON. I don't know what you get in the case of authentication error. Perhaps that's the reason you can't catch the authentication errors?

src/server_manager/cloud/digitalocean_api.ts
fortuna, line 224: This will trigger an infinite retry in the case of authentication error.

src/server_manager/web_app/digitalocean_app/create_server_app.ts
```diff
@@ -0,0 +1,137 @@
...
+  private static EVENT_CREATE_SERVER_REQUESTED = 'DigitalOceanCreateServerApp#_RequestCreateServer';
+
+  @property({type: Function}) localize: Function;
+
+  private notificationManager: OutlineNotificationManager;
+  private eventEmitter = new EventEmitter();
+
+  constructor() {
+    super();
+    this.addEventListener(OutlineRegionPicker.EVENT_REGION_SELECTED, (event: CustomEvent) => {
```
fortuna, line 55: You don't need to call addEventListener. You can specify the listener on the render template instead.

src/server_manager/web_app/ui_components/app-root.js
```diff
@@ -655,48 +643,23 @@ export class AppRoot extends mixinBehaviors
...
-    this.$.toast.close();
+  getNotificationManager() {
+    return this.$.notificationManager;
   }
 
   /**
-   * @param {(retry: boolean) => void} cb a function which accepts a single boolean which is true
-   *     iff
-   *      the user chose to retry the failing operation.
+   *
```
fortuna, line 654: Empty comment. Perhaps add the return type?

src/server_manager/web_app/ui_components/outline-notification-manager.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  showErrorRaw(message: string) {
+    this.showToast(message, Infinity);
+  }
+
+  showNotification(messageId: string, durationMs = 3000) {
+    this.showToast(this.localize(messageId), durationMs);
+  }
+
+  private showToast(message: string, duration: number) {
```
fortuna, line 68: Please rename `duration` to indicate the unit (`durationSeconds`?)

src/server_manager/web_app/ui_components/outline-region-picker-step.ts
```diff
@@ -110,12 +113,14 @@ export class OutlineRegionPicker extends LitElement {
       <span slot="step-title">${this.localize('region-title')}</span>
       <span slot="step-description">${this.localize('region-description')}</span>
       <span slot="step-action">
-        <paper-button id="createServerButton" @tap="${this._handleCreateServerTap}" ?disabled="${!this._isCreateButtonEnabled(this.isServerBeingCreated, this.selectedLocationId)}">
+        <paper-button id="createServerButton" @tap="${this._handleCreateServerTap}" ?disabled="${
+        !this._isCreateButtonEnabled(this.isServerBeingCreated, this.selectedLocationId)}">
```
fortuna, line 117: I think that inlining the logic here will make the code more readable.

src/server_manager/web_app/ui_components/outline-region-picker-step.ts
```diff
@@ -148,12 +156,8 @@ export class OutlineRegionPicker extends LitElement {
 
   _handleCreateServerTap(): void {
     this.isServerBeingCreated = true;
```
fortuna, line 158: That's application logic. Let's make the parent set this instead.

src/server_manager/web_app/digitalocean_app/create_server_app.ts
fortuna, line 55: Why do you need to forward this event? Call a method to handle the creation directly instead.

src/server_manager/web_app/digitalocean_app/create_server_app.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  setNotificationManager(notificationManager: OutlineNotificationManager) {
+    this.notificationManager = notificationManager;
+  }
+
+  async start(digitalOceanServerRepository: ManagedServerRepository): Promise<void> {
+    await this.showRegionPicker(digitalOceanServerRepository);
+
+    this.eventEmitter.once(
```
fortuna, line 73: This is not needed. All you need is to have a handler for EVENT_REGION_SELECTED that is set one in the render call.

I also don't understand why you need sever created and create failure events. Who handles them? This app is supposed to be self-contained.

The only event the parent cares about is a "ServerCreated(Server)" event, which may be emitted by this app, or by the repository. We can make the app emit it for now to keep things a bit simpler.

fortuna (changes requested):
This is looking good. I think we can simplify all the event handling in the create server app.

src/server_manager/web_app/ui_components/outline-region-picker-step.ts
mpmcroy: Done

src/server_manager/web_app/digitalocean_app/create_server_app.ts
```diff
@@ -38,58 +37,50 @@ const DIGITALOCEAN_FLAG_MAPPING: {[cityId: string]: string} = {
...
+    try {
+      const server =
+          await this.createServer(this.digitalOceanServerRepository, event.detail.regionId);
+      const serverCreatedEvent =
+          makePublicEvent(DigitalOceanCreateServerApp.EVENT_SERVER_CREATED, {server});
+      this.dispatchEvent(serverCreatedEvent);
+    } catch (error) {
+      this.notificationManager.showError('error-server-creation');
+    }
   }
```
fortuna, line 68: I think we should set `regionPicker.isServerBeingCreated = false` in a finally block here

src/server_manager/web_app/app.ts
```diff
@@ -342,7 +332,8 @@ export class App {
         messageKey = 'error-servers-removed';
         placeholder = 'serverNames';
       }
-      this.appRoot.showError(this.appRoot.localize(messageKey, placeholder, unsyncedServerNames));
+      this.notificationManager.showErrorRaw(
```
fortuna, line 335: Do you need showErrorRaw?
You could make the showError signature match the localize signature and take parameters.

src/server_manager/web_app/digitalocean_app/create_server_app.ts
```diff
@@ -0,0 +1,128 @@
...
+          await this.createServer(this.digitalOceanServerRepository, event.detail.regionId);
+      const serverCreatedEvent =
+          makePublicEvent(DigitalOceanCreateServerApp.EVENT_SERVER_CREATED, {server});
+      this.dispatchEvent(serverCreatedEvent);
+    } catch (error) {
+      this.notificationManager.showError('error-server-creation');
+    }
+  }
+
+  setNotificationManager(notificationManager: OutlineNotificationManager) {
```
fortuna, line 70: Do we need an explicit setter? Perhaps just set the property?

src/server_manager/web_app/digitalocean_app/create_server_app.ts
```diff
@@ -0,0 +1,128 @@
...
+  async start(digitalOceanServerRepository: ManagedServerRepository): Promise<void> {
+    this.digitalOceanServerRepository = digitalOceanServerRepository;
+    return this.showRegionPicker(this.digitalOceanServerRepository);
+  }
+
+  private async showRegionPicker(digitalOceanServerRepository: ManagedServerRepository):
+      Promise<void> {
+    const regionPicker =
+        this.shadowRoot.querySelector('outline-region-picker-step') as OutlineRegionPicker;
+    regionPicker.isServerBeingCreated = false;
```
fortuna, line 83: reset should set isServerBeingCreated, so no need for this line

src/server_manager/web_app/digitalocean_app/create_server_app.ts
```diff
@@ -0,0 +1,128 @@
...
+      const locations = Object.entries(regionMap).map(([cityId, regionIds]) => {
+        return this.createLocationModel(cityId, regionIds);
+      });
+      regionPicker.locations = locations;
+    } catch (error) {
+      if (error instanceof XhrError) {
+        const event = makePublicEvent(DigitalOceanCreateServerApp.EVENT_AUTHORIZATION_ERROR);
+        this.dispatchEvent(event);
+      } else {
+        console.error(`Failed to get list of available regions: ${error}`);
```
fortuna, line 98: You should still show an error in the case of authentication error.

src/server_manager/web_app/digitalocean_app/create_server_app.ts
```diff
@@ -0,0 +1,128 @@
...
+    try {
+      // TODO: Catch and rethrow authorization error
+      const regionMap = await digitalOceanServerRepository.getRegionMap();
+      const locations = Object.entries(regionMap).map(([cityId, regionIds]) => {
+        return this.createLocationModel(cityId, regionIds);
+      });
+      regionPicker.locations = locations;
+    } catch (error) {
+      if (error instanceof XhrError) {
+        const event = makePublicEvent(DigitalOceanCreateServerApp.EVENT_AUTHORIZATION_ERROR);
```
fortuna, line 95: Catching authorization errors and bubbling up everywhere you need is error-prone. Each app will have to have their own handling.

I think it's better if the repo emits an event on authorization error, and have the root app listen on that. This way handling happens in one place only.

For this app, simply show the error you have below. Don't differentiate the errors.

src/server_manager/web_app/digitalocean_app/create_server_app.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    try {
+      // TODO: Catch and rethrow authorization error
+      const regionMap = await digitalOceanServerRepository.getRegionMap();
+      const locations = Object.entries(regionMap).map(([cityId, regionIds]) => {
+        return this.createLocationModel(cityId, regionIds);
+      });
+      regionPicker.locations = locations;
+    } catch (error) {
+      if (error instanceof XhrError) {
```
fortuna, line 94: This doesn't read right. Why is a XhrError an authorization error? Could you make this more readable?

Also, don't you have to check the status code?

src/server_manager/web_app/digitalocean_app/create_server_app.ts
```diff
@@ -0,0 +1,128 @@
...
+        const event = makePublicEvent(DigitalOceanCreateServerApp.EVENT_AUTHORIZATION_ERROR);
+        this.dispatchEvent(event);
+      } else {
+        console.error(`Failed to get list of available regions: ${error}`);
+        this.notificationManager.showError('error-do-regions');
+      }
+    }
+  }
+
+  private createServer(digitalOceanServerRepository: ManagedServerRepository, regionId: string):
```
fortuna, line 104: I don't see much use for this method. Perhaps inline it instead?

src/server_manager/web_app/digitalocean_app/create_server_app.ts
```diff
@@ -0,0 +1,128 @@
...
+};
+
+@customElement('digitalocean-create-server-app')
+export class DigitalOceanCreateServerApp extends LitElement {
+  public static EVENT_SERVER_CREATED = 'do-create-server-app-server-created';
+  public static EVENT_AUTHORIZATION_ERROR = 'do-create-server-app-authorization-error';
+
+  @property({type: Function}) localize: Function;
+
+  private digitalOceanServerRepository: ManagedServerRepository;
```
fortuna, line 45: Perhaps digitalOceanServerRepository -> serverRepository? This app is already in the DO scope.

src/server_manager/web_app/ui_components/outline-notification-manager.ts
```diff
@@ -0,0 +1,87 @@
...
+        <paper-toast id="toast">
+            <paper-icon-button icon="icons:close" @tap="${this.closeToast}"></paper-icon-button>
+        </paper-toast>`;
+  }
+
+  showError(messageId: string) {
+    this.showToast(this.localize(messageId), Infinity);
+  }
+
+  showErrorRaw(message: string) {
```
fortuna, line 60: I don't think this method is needed. You could change showError to take ...args, and pass that to the localize call.

src/server_manager/web_app/ui_components/outline-notification-manager.ts
```diff
@@ -0,0 +1,87 @@
...
+    this.showToast(message, Infinity);
+  }
+
+  showNotification(messageId: string, durationMs = 3000) {
```
fortuna, line 64: You are assigning durationMs to durationSeconds. There's a unit mismatch there.

src/server_manager/web_app/ui_components/outline-region-picker-step.ts
```diff
@@ -32,6 +32,8 @@ export interface Location {
 
 @customElement('outline-region-picker-step')
 export class OutlineRegionPicker extends LitElement {
+  public static EVENT_REGION_SELECTED = 'outline-region-picker-region-selected';
```
fortuna, line 35: Let's shorten this to region-selected

src/server_manager/web_app/ui_components/outline-region-picker-step.ts
```diff
@@ -127,33 +131,26 @@ export class OutlineRegionPicker extends LitElement {
...
 
   _handleCreateServerTap(): void {
-    this.isServerBeingCreated = true;
-    const params = {
-      bubbles: true,
-      composed: true,
-      detail: {selectedRegionId: this.selectedLocationId}
-    };
-    const customEvent = new CustomEvent('RegionSelected', params);
+    const params = {bubbles: true, composed: true, detail: {regionId: this.selectedLocationId}};
```
fortuna, line 152: Perhaps we don't need those parameters, since you are adding the event listener to the element directly.
Could you try using an Event without those parameters?

fortuna (changes requested):
This is looking good. But I'm afraid the handling of authorization errors is still a bit fragile and complex.

I suggest we make the repo emit an unauthorized event that the root app listens to directly, so the different components don't need to handle it.

src/server_manager/web_app/digitalocean_app/create_server_app.ts
```diff
@@ -0,0 +1,271 @@
...
+        width: 100%;
+        border: 1px solid rgba(255, 255, 255, 0.12);
+        border-radius: 2px;
+      }
+      .card paper-button[disabled] {
+        color: var(--medium-gray);
+        background: transparent;
+      }
+      /* Mirror images */
+      :host(:dir(rtl)) .mirror {
```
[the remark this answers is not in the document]
fortuna: Perhaps you can reproduce a bit of the [DirMixin logic](https://github.com/Polymer/polymer/blob/master/lib/mixins/dir-mixin.js)

What you can do i set the host `dir` attribute on connected, based on the `<html>` `dir` attribute, if the host attribute is not already set. See https://github.com/Polymer/polymer/blob/master/lib/mixins/dir-mixin.js#L176

Ideally it would be inherited, but I don't know how to do that.

Once you have the `dir` attribute, you can do
```css
:host([dir="rtl"]) .mirror {
}
```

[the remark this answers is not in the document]
fortuna: Perhaps you can use this to calculate the inherited direction:
```js
this.setAttribute("dir", getComputedStyle(this).direction())
```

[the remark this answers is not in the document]
fortuna: cc: @JonathanDCohen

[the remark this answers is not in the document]
fortuna: The gallery app may be a great place to play with that.

fortuna (changes requested):

PR #732, opened 2020-09-17
Remove unused assets and code

JonathanDCohen:
This removes the `empty_user.png` and `server_avatar.png` images.  

This also removes the `LoadingWindow` class.  We make one in index.ts but never actually open the window.

Tested with code searching and running the app.


alalamav:
I think we may have accidentally deleted the code that sets a timer to display the loading window. I think there's value in having it as a fallback the main window takes too long or fails to load. Maybe call `loadingWindow.showInMs(LOADING_WINDOW_DELAY_MS)`?

JonathanDCohen:
I removed the change to loadingwindow to put in a different PR.  All else good?

alalamav (approved):

PR #736, opened 2020-09-24
Release channels and staged releases for the manager

JonathanDCohen:
This PR allows us to release a beta or alpha build of the manager by appending "-beta" or "-alpha" to the version field in the manager's `package.json`. It also allows us to perform a staged rollout by appending `-s $rollout_percentage` to the build command.  This is similar to https://github.com/Jigsaw-Code/outline-client/pull/841 for the client.

This PR adds [jq](https://stedolan.github.io/jq/) to parse `package.json` as a system dependency requirement for packaging or releasing the manager in order to parse `package.json`


README.md
```diff
@@ -46,6 +46,8 @@ Note: If you are using root (not recommended on your dev machine, maybe in a con
 
 This project uses [Yarn workspaces](https://yarnpkg.com/blog/2017/08/02/introducing-workspaces/).
 
+In order to package or build a release version of the Manager, you will need [jq](https://stedolan.github.io/jq/) to parse `package.json`
```
alalamav, line 49: Let's use [jq-node](https://www.npmjs.com/package/node-jq) as a dev dependency so it gets installed with yarn.

src/server_manager/electron_app/release_macos_action.sh
```diff
@@ -21,16 +21,48 @@ if [ "$CSC_LINK" == "unset" ]; then
   exit 1
 fi
 
+function usage () {
```
alalamav, line 24: nit: can we can we move this code to a separate script (i.e. `get_packaging_opts.sh`) and call it from all the scripts?

src/server_manager/scripts/manager_release_channel.sh
```diff
@@ -0,0 +1,8 @@
+#! /bin/bash
```
alalamav, line 1: nit: copyright

README.md
JonathanDCohen: Done in 5a647d9abf6896c01e85dbb8f1343f522bf2fee6

JonathanDCohen (commented):
PTAL.  Note that I fixed a bug regarding staged releases of mainline releases: Beta channel users take updates from both the beta and the mainline channels.  Therefore if we cut a staged rollout on the mainline we need to do the same for the beta channel.

src/server_manager/scripts/get_packaging_opts.sh
```diff
@@ -0,0 +1,54 @@
...
+done
+
+if ((STAGING_PERCENTAGE <= 0)) || ((STAGING_PERCENTAGE > 100)); then
+  echo "Staging percentage must be greater than 0 and no more than 100" 1>&2
+  exit 1
+fi
+
+# This breaks if we have more than one option.  We could, for example, use jq here to return JSON
+# and jq from the calling script to parse it to variables.
+echo "${STAGING_PERCENTAGE}"
```
alalamav, line 54: An alternative would be to have the parent process export `$STAGING_PERCENTAGE`, along with any other options, and have this script populate them accordingly. It may be worth implementing this contract now to avoid the TODO - up to you.

src/server_manager/scripts/manager_release_channel.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash -eu
```
alalamav, line 1: nit: get_manager_release_channel?

alalamav (approved):

src/server_manager/scripts/get_packaging_opts.sh
JonathanDCohen: This doesn't quite work as you propose, since subprocesses get a copy of the parent process' environment.  I did some research, however, and found that scripts called with `source` run in the same process as the caller, and so can populate variables without them even needing to be exported to the environment at all!  Thanks for the motivation to find a better solution \^_\^

Done in  3ba6a3c

JonathanDCohen (commented):
renaming to get_manager_release_channel done in 3401fb7

src/server_manager/scripts/get_packaging_opts.sh
alalamav: Glad my suggestion helped 😅

alalamav (approved):
Thanks for the changes 👍

PR #757, opened 2020-10-28
Apply postcss processor on LitElement css

mpmcroy:
**Important: It seems like LitElement isn't recognizing the `:dir` selector. So although we're generating the LTR/RTL css, neither is being applied. We may need to apply something similar to the Polymer DirMixin.**

This extends the CSS RTL loader to process LitElement css blocks (similar to the way we process `<style>` tags in Polymer elements).

Below is an example of how the `outline-about-dialog` css is transformed:

**Before**
```
p {
  margin: 20px 0 0 0;
  text-align: left;
}
```

**After**
```
p {
  margin: 20px 0 0 0;
}
:host(:dir(ltr)) p {
  text-align: left;
}
:host(:dir(rtl)) p {
  text-align: right;
}
```


JonathanDCohen:
Also worth noting https://github.com/Polymer/lit-element/issues/633, that you can't use the `styles` property for css with custom property mixins, which includes anything to style our paper elements.

PR #768, opened 2020-11-18
Cohenjon per key limits shadowbox

JonathanDCohen:


src/shadowbox/server/server_access_key.ts
```diff
@@ -181,19 +179,28 @@ export class ServerAccessKeyRepository implements AccessKeyRepository {
   }
 
   get dataLimit(): DataLimit|undefined {
-    return this.accessKeyDataLimit;
+    return this.defaultDataLimit;
   }
 
-  setAccessKeyDataLimit(limit: DataLimit): Promise<void> {
```
JonathanDCohen, line 187: This validation was removed because it duplicates validation done at the API boundary.  By the time we're here, we don't have an unknown JSON request but a known valid DataLimit.

src/shadowbox/server/manager_service.ts
```diff
@@ -28,7 +28,7 @@ import {ServerConfigJson} from './server_config';
 import {SharedMetricsPublisher} from './shared_metrics';
 
 // Creates a AccessKey response.
-function accessKeyToJson(accessKey: AccessKey) {
+function accessKeyResponseJson(accessKey: AccessKey) {
```
JonathanDCohen, line 31: This rename goes along with the above rename of MakeAccessKeyJson to accessKeySerializedJson in order to better disambiguate the two different "Access Key to JSON" functions.

PR #771, opened 2020-11-21
Change "access key data limit" to "default data limit" in the manager

JonathanDCohen:
This is a purely no-op change to refer to "server default" data limits instead of "access key data limits".  Doing this as a separate change before adding per-key data limits to the manager will help keep the diff contained to functional changes.

This PR is to merge to `cohenjon-manager-base`, which will be rebased to the `cohenjon-per-key-limits-shadowbox` branch as necessary until that branch is merged into master, in order to be able to independently develop shadowbox and the manager in two different pull requests.

Tested:
yarn do server_manager/test
built the electron app, set and removed default limits, changed a default limit, made sure all data limit UI worked correctly.


alalamav (approved):
Thanks for breaking up the change, LGTM.

PR #782, opened 2020-12-23
fix wrong path in readme

newjett0617:
as title


newjett0617:
@googlebot I signed it!

fortuna (approved):
Thanks for the fix!

PR #784, opened 2020-12-23
Chore: upgrade deps.  Also upgrades to Typescript 4

JonathanDCohen:
This supersedes a number of dependabot security PRs.

Generated by changing package.json and running `yarn upgrade --caret`

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "husky": "^1.3.1",
     "jasmine": "^3.5.0",
     "tslint": "^5.9.1",
-    "typescript": "^3.8.3"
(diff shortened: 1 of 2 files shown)


fortuna (approved):

fortuna (commented):
The tests are failing. Perhaps some dependency update breaks us?

fortuna (changes requested):
I see an electron-related error in the unit test:
image

You may need to revert some dependencies. I've had issues upgrading Electron-related dependencies in the past too.

fortuna (commented):
It could be Sentry too.

PR #795, opened 2021-01-14
Adds 'serverId' to all events that affect the server

mpmcroy:
This updates the events that affect the server to include the server ID. This allows us to get rid of selected server local state in `app.ts` (`selectedServer`).

## Open questions
* The `ShadowboxServer`/`DigitalOceanServer`/`ManualServer` hierarchy is a little confusing now that ShadowboxServer is abstract.
* Determine what we should name the [managed/manual] server ID (what we're calling localServerId atm) vs the shadowbox Server ID

### DIFF
--- a/src/server_manager/web_app/app.spec.ts
+++ b/src/server_manager/web_app/app.spec.ts
@@ -18,7 +18,7 @@ import * as digitalocean_api from '../cloud/digitalocean_api';
 import {sleep} from '../infrastructure/sleep';
 import * as server from '../model/server';
 
-import {App, LAST_DISPLAYED_SERVER_STORAGE_KEY, localServerId} from './app';
+import {App, LAST_DISPLAYED_SERVER_STORAGE_KEY} from './app';
 import {TokenManager} from './digitalocean_oauth';
 import {AppRoot} from './ui_components/app-root.js';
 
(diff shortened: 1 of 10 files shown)


src/server_manager/web_app/app.spec.ts
```diff
@@ -193,7 +194,7 @@ class FakeServer implements server.Server {
     this.metricsEnabled = metricsEnabled;
     return Promise.resolve();
   }
-  getServerId() {
+  getMetricsId() {
```
mpmcroy, line 197: This is the shadowboxServerId – so we should come up with a better name. Maybe we revert it back to serverId.

We also need to think through what the names of these are in the views. We should not use `id` so that we don't collide with the HTML id attribute.

src/server_manager/web_app/app.ts
```diff
@@ -180,39 +171,44 @@ export class App {
     });
 
     appRoot.addEventListener('DeleteServerRequested', (event: CustomEvent) => {
-      this.deleteSelectedServer();
+      this.deleteSelectedServer(event.detail.serverId);
```
mpmcroy, line 174: TODO: Rename to deleteServer

src/server_manager/web_app/app.ts
```diff
@@ -180,39 +171,44 @@ export class App {
...
     });
 
     appRoot.addEventListener('ForgetServerRequested', (event: CustomEvent) => {
-      this.forgetSelectedServer();
+      this.forgetSelectedServer(event.detail.serverId);
```
mpmcroy, line 178: TODO: Rename to forgetServer

PR #796, opened 2021-01-14
Internationalize the label for the default data limit input

JonathanDCohen:


fortuna (approved):
Is the label from mocks? You may want to check with Monique or Justin if the wording is ok.

JonathanDCohen:
It's a no-op in English.

Issue #774, opened 2020-11-30
[outline-shadowsocksconfig] `userinfo` in generated ss:// links shouldn't include padding characters


database64128:
This issue was intended for [Jigsaw-Code/outline-shadowsocksconfig](https://github.com/Jigsaw-Code/outline-shadowsocksconfig). But since the repo has the issues area disabled, I have no choice but to post it here.

[SIP002](https://shadowsocks.org/en/wiki/SIP002-URI-Scheme.html) requires the `userinfo` to be `base64url`-encoded. That means the trailing padding characters (`=`) shouldn't be included. For example, instead of `ss://YWVzLTI1Ni1nY206cGFzc3dvcmQ=@github.com:443/#GitHub`, the link should be `ss://YWVzLTI1Ni1nY206cGFzc3dvcmQ@github.com:443/#GitHub`.

Note that previously the standard document has some incorrect examples. I have fixed them in [this commit](https://github.com/shadowsocks/shadowsocks-org/wiki/SIP002-URI-Scheme/_compare/dc364be339684947d3778193166768f474bb43ea...73c73295e7bb256b0c80975db16c152405dff05d).

ohnorobo:
This still needs a release to be cut for for outline-shadowsocksconfig

ohnorobo:
cut a release https://github.com/Jigsaw-Code/outline-shadowsocksconfig/releases/tag/v0.1.1

database64128:
Thank you for taking time to address the issue! 👍

ohnorobo:
Also updated in https://github.com/Jigsaw-Code/outline-client/pull/965

PR #815, opened 2021-02-11
bump ssconfig version

ohnorobo:
To match newly cut version https://github.com/Jigsaw-Code/outline-shadowsocksconfig/releases/tag/v0.1.2

We never actually upgraded outlline-server to use the previous release, v.0.1.0. (Although outline client has been upgraded for a while.) So this is pulling in several config changes and lib updates: https://github.com/Jigsaw-Code/outline-shadowsocksconfig/compare/v0.0.9...v0.1.2

This is to fix https://github.com/Jigsaw-Code/outline-server/issues/774

### DIFF
--- a/src/shadowbox/package.json
+++ b/src/shadowbox/package.json
@@ -10,7 +10,7 @@
     "Using https:// for ShadowsocksConfig to avoid adding git in the Docker image"
   ],
   "dependencies": {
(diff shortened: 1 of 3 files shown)


alalamav (approved):
Good catch, thank you for updating this.

PR #822, opened 2021-02-16
Require scripts to be shellcheck-compliant

bemasc:
This also adds shellcheck verification to the presubmit lint step.


src/metrics_server/run_action.sh
```diff
@@ -14,12 +14,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-SRC_DIR="src/metrics_server"
-BUILD_DIR="build/metrics_server"
+readonly SRC_DIR="src/metrics_server"
```
fortuna, line 17: When should we use `readonly` and when `local -r`? You've replaced the `readonly` in the other file.
Do we really need different syntaxes based on whether it's a function or script-scope variable?

src/server_manager/electron_app/package_linux_action.sh
```diff
@@ -32,4 +33,4 @@ $ROOT_DIR/src/server_manager/node_modules/.bin/electron-builder \
   --config.linux.category=Network \
   --config.artifactName='Outline-Manager.${ext}'
```
fortuna, line 34: I wonder if ` --config.artifactName="Outline-Manager.\${ext}"` is a better way to remove the warning.

src/shadowbox/server/run_action.sh
```diff
@@ -16,21 +16,23 @@
...
-readonly STATE_CONFIG=${SB_STATE_DIR}/shadowbox_server_config.json
+readonly STATE_CONFIG="${SB_STATE_DIR}/shadowbox_server_config.json"
 
-[[ -d "${SB_STATE_DIR}" ]] || mkdir -p $SB_STATE_DIR
-[[ -e $STATE_CONFIG ]] || echo '{"hostname":"127.0.0.1"}' > $STATE_CONFIG
+[[ -d "${SB_STATE_DIR}" ]] || mkdir -p "${SB_STATE_DIR}"
+[[ -e "${STATE_CONFIG}" ]] || echo '{"hostname":"127.0.0.1"}' > "${STATE_CONFIG}"
 
-source $ROOT_DIR/src/shadowbox/scripts/make_test_certificate.sh $RUN_DIR
+# shellcheck source=src/shadowbox/scripts/make_test_certificate.sh
```
fortuna, line 35: What does this do?
Shouldn't the path be relative?

package.json
```diff
@@ -16,7 +16,8 @@
   "scripts": {
     "clean": "rm -rf src/*/node_modules/ build/ node_modules/ src/server_manager/install_scripts/do_install_script.ts",
     "do": "bash ./scripts/do_action.sh",
-    "lint": "tslint 'src/**/*.ts' -e '**/node_modules/**'"
+    "lint": "tslint 'src/**/*.ts' -e '**/node_modules/**'",
```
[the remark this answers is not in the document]
fortuna: This is annoying. Can we download the binary and put it somewhere?

Something like this could be in our package.json dev install, perhaps in a script:
```bash
mkdir -p ./downloads/
curl --silent -L "https://github.com/koalaman/shellcheck/releases/download/latest/shellcheck-latest.$(uname | tr [:upper:] [:lower:]).x86_64.tar.xz" | tar x --directory ./downloads
```

I'm using `downloads/` as a suggestion. It could be in `node_modules/`, but that may upset yarn verification.

fortuna (commented):
Looks good, but perhaps we can fix the need to install shellcheck

package.json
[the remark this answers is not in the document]
fortuna: Actually, I think it's better if we pin to a version of shellcheck. Like https://github.com/koalaman/shellcheck/releases/download/v0.7.1/shellcheck-v0.7.1.${PLATFORM}.x86_64.tar.xz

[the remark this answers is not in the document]
bemasc: OK, done.  This was nontrivial, due to the need to handle various platforms, avoid repeatedly downloading the binary, check hashes, etc.

In the process of implementing this I noticed that our "husky" pre-commit hook is not running (at least not for me) so we are not actually running the linters on every commit.

third_party/shellcheck/run.sh
```diff
@@ -0,0 +1,50 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+readonly VERSION='v0.7.1'
+readonly DOWNLOAD='download'
+
+# The relative location of this script.
+dir="$(dirname "$0")/${DOWNLOAD}"
+readonly dir
```
fortuna, line 22:
```suggestion
# The relative location of this script.
readonly DOWNLOAD_DIR="$(dirname "$0")/${DOWNLOAD}"
```

The `DOWNLOAD` variable is not really useful and hurts readability.

fortuna (approved):
Thanks for the changes

package.json
```diff
@@ -14,9 +14,11 @@
     "node": "^12"
   },
   "scripts": {
-    "clean": "rm -rf src/*/node_modules/ build/ node_modules/ src/server_manager/install_scripts/do_install_script.ts",
+    "clean": "rm -rf src/*/node_modules/ build/ node_modules/ src/server_manager/install_scripts/do_install_script.ts third_party/shellcheck/download/",
     "do": "bash ./scripts/do_action.sh",
-    "lint": "tslint 'src/**/*.ts' -e '**/node_modules/**'"
+    "lint": "yarn shellcheck && yarn tslint",
+    "shellcheck": "find src scripts third_party/shellcheck -name *.sh -exec third_party/shellcheck/run.sh {} +",
```
JonathanDCohen, line 20: When I run on cygwin this finds errors in src/server_manager/bower_components/intl-messageformat/tests/saucelabs.sh.  Maybe we can ignore src/server_manager/bower_components?

third_party/shellcheck/run.sh
bemasc: Done

package.json
bemasc: OK, I changed this to ignore `bower_components`.  Could you try again?

This made the `find` invocation too complex to be in a string in `package.json`, so I moved it into `scripts/shellcheck.sh`.

package.json
```diff
@@ -14,9 +14,11 @@
     "node": "^12"
   },
   "scripts": {
-    "clean": "rm -rf src/*/node_modules/ build/ node_modules/ src/server_manager/install_scripts/do_install_script.ts",
+    "clean": "rm -rf src/*/node_modules/ build/ node_modules/ src/server_manager/install_scripts/do_install_script.ts third_party/shellcheck/download/",
     "do": "bash ./scripts/do_action.sh",
-    "lint": "tslint 'src/**/*.ts' -e '**/node_modules/**'"
+    "lint": "yarn shellcheck && yarn tslint",
+    "shellcheck": "bash ./scripts/shellcheck.sh",
+    "tslint": "tslint 'src/**/*.ts' -e '**/node_modules/**'"
```
JonathanDCohen, line 21: The shellcheck change worked, but now we're triggering tslint on bower_components as well.

```suggestion
    "tslint": "tslint 'src/**/*.ts' -e '**/node_modules/**' -e '**/bower_components/**'"
```

JonathanDCohen (approved):
Looks good except for the tslint suggestion.

package.json
fortuna: @JonathanDCohen I think you have an unclean client. We no longer use bower.

PR #823, opened 2021-02-17
Slightly nontrivial shell cleanup

bemasc:
These are slightly nontrivial shell cleanups discovered in the course of
attempting shellcheck compliance.


fortuna (approved):
Thanks for the cleanup. We may want to test this again.

bemasc:
I've tested all of the changed scripts and they seem to work normally.

PR #825, opened 2021-02-18
Update sentry webhook runtime to node 12

alalamav:
- The sentry webhook cloud function is running on a deprecated node 6 runtime, causing all requests to fail.
- Updates the runtime to node 12.


fortuna (approved):
Thanks for taking care of this

alalamav:
I deployed the cloud function and verified that it is posting to Salesforce.

PR #832, opened 2021-02-23
Integrates the account model with the UI

mpmcroy:
First stab at integrating the account model with the UI.

This is still very DigitalOcean specific atm.


src/server_manager/web_app/app.ts
```diff
@@ -364,15 +364,28 @@ export class App {
...
     return {
       id: server.getId(),
+      accountId: account ? this.makeGloballyUniqueAccountId(account) : null,
       name: this.makeDisplayName(server),
-      isManaged: isManagedServer(server),
       isSynced: !!server.getName(),
     };
   }
 
+  private makeGloballyUniqueAccountId(account: digitalocean.Account): string {
```
mpmcroy, line 385: This is generated solely for the UI. We don't persist this anywhere.

PR #833, opened 2021-02-24
Fix shellcheck installation on macOS

bemasc:
The shellcheck download verification step was relying on `sha256sum`,
which is not present in the macOS default install.


mpmcroy (approved):
I only tested on Mac, but it works great. Thanks!

PR #839, opened 2021-03-04
Rename adminEmail to digitalOceanAccountName

mpmcroy:
Renames the `app-root.js` DigitalOcean `adminEmail` to `digitalOceanAccountName` and simplifies `outline-intro-step` by removing `isSignedInToDigitalOcean`.

Unfortunately we still need the `_compute*` methods to signal to the Polymer code that an account is connected.


fortuna (approved):
Nice little cleanup.

By the way, if accounts was a list, then you wouldn't need the hidden parameter. You'd have a dom-repeat instead.

PR #778, opened 2020-12-15
UI for Per Key Data Limits

JonathanDCohen:
[Demo Video of setting and removing data limits](https://youtu.be/9g8ZzinV_5I)

[Demo showing enforcement of data limit](https://youtu.be/eG2e1uC2LpQ)

This PR adds the option to set a data limit for each access key in the Manager app.  This limit overrides the server default data limit if it exists, but can also be set even if no server default data limit is set.  

The limit is able to be set through a new dialog window, which is opened and closed either via the overflow menu for user access keys or from a dedicated button on the admin access key row.  This may change to an overflow menu when more features are added.   This  dialog box saves no state when closed and relies on the appRoot to accurately open it with the correct key and server data limit.  Saving the data limit requires hitting a Save button in the dialog.

 If the dialog is opened and nothing is changed, hitting the "Save" button won't send any requests to the server.

PR #841, opened 2021-03-09
Per-Key Data Limits in the Manager

JonathanDCohen:
This supercedes https://github.com/Jigsaw-Code/outline-server/pull/778 in order to get around a complicated git history.  Please refer there for further discussion.

https://user-images.githubusercontent.com/3297872/110400206-3a0d9700-8045-11eb-887a-49886d89c9da.mov


src/server_manager/web_app/app.ts
```diff
@@ -855,42 +858,82 @@ export class App {
...
       return;
     }
     const serverView = this.appRoot.getServerView(this.appRoot.selectedServerId);
     try {
-      await this.selectedServer.setAccessKeyDataLimit(limit);
+      await this.selectedServer.setDefaultDataLimit(limit);
       this.appRoot.showNotification(this.appRoot.localize('saved'));
-      serverView.accessKeyDataLimit = dataLimitToDisplayDataAmount(limit);
+      serverView.defaultDataLimitBytes = limit?.bytes;
+      serverView.isDefaultDataLimitEnabled = !!limit;
```
fortuna, line 874: This is always true because you return early if the limit is absent.
```suggestion
      serverView.isDefaultDataLimitEnabled = true;
```

src/server_manager/web_app/app.ts
```diff
@@ -855,42 +858,82 @@ export class App {
...
 
-  private async removeAccessKeyDataLimit() {
+  private async removeDefaultDataLimit() {
     const serverView = this.appRoot.getServerView(this.appRoot.selectedServerId);
     try {
-      await this.selectedServer.removeAccessKeyDataLimit();
+      await this.selectedServer.removeDefaultDataLimit();
       this.appRoot.showNotification(this.appRoot.localize('saved'));
       this.refreshTransferStats(this.selectedServer, serverView);
     } catch (error) {
```
fortuna, line 894: Don't you need to set `serverView.isDefaultDataLimitEnabled = false` here?

src/server_manager/web_app/app.ts
```diff
@@ -855,42 +858,82 @@ export class App {
...
-      console.error(`Failed to remove access key data limit: ${error}`);
+      console.error(`Failed to remove server default data limit: ${error}`);
       this.appRoot.showError(this.appRoot.localize('error-remove-data-limit'));
-      serverView.isAccessKeyDataLimitEnabled = true;
+      serverView.isDefaultDataLimitEnabled = true;
```
fortuna, line 897: I think this is not needed and may break. What if you call remove twice?

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
```diff
@@ -0,0 +1,380 @@
...
+  }
+
+  private renderMenu() {
+    return html`
+      <div id="menu">
+        <paper-input
+          id="dataLimitInput"
+          label=${this.localize('data-limit')}
+          always-float-label
+          allowed-pattern="[0-9\\.]"
```
fortuna, line 204: I wonder how well this internationalized.
Different countries use different number separators.

src/server_manager/web_app/app.ts
```diff
@@ -855,42 +858,82 @@ export class App {
...
+      if (dialog.dataLimitChanged()) {
+        const dataLimit = dialog.inputDataLimit();
+        await server.setAccessKeyDataLimit(keyId, displayDataAmountToDataLimit(dataLimit));
+        this.refreshTransferStats(server, serverView);
+      }
+      dialog.close();
+      this.appRoot.showNotification(this.appRoot.localize('saved'));
+    } catch (error) {
+      console.error(`Failed to set data limit for access key ${keyId}: ${error}`);
+      dialog.setInitialMenuState();
```
fortuna, line 916: Let's not change the state, so the user can see what was there.

src/server_manager/web_app/app.ts
```diff
@@ -855,42 +858,82 @@ export class App {
...
+    try {
+      if (dialog.dataLimitChanged()) {
+        await server.removeAccessKeyDataLimit(keyId);
+        this.refreshTransferStats(server, serverView);
+      }
+      dialog.close();
+      this.appRoot.showNotification(this.appRoot.localize('saved'));
+    } catch (error) {
+      console.error(`Failed to remove data limit from access key ${keyId}: ${error}`);
+      dialog.setInitialMenuState();
```
fortuna, line 935: Same

src/server_manager/web_app/ui_components/outline-server-settings.js
```diff
@@ -208,7 +208,7 @@ Polymer({
               <!-- NOTE: The dropdown is not automatically sized to the button's width:
                            https://github.com/PolymerElements/paper-dropdown-menu/issues/229 -->
               <paper-dropdown-menu no-label-float="" horizontal-align="left">
-                <paper-listbox slot="dropdown-content" selected="{{_computeDataLimitsEnabledName(isAccessKeyDataLimitEnabled)}}" attr-for-selected="name" on-selected-changed="_accessKeyDataLimitEnabledChanged">
+                <paper-listbox slot="dropdown-content" selected="{{_computeDataLimitsEnabledName(isDefaultDataLimitEnabled)}}" attr-for-selected="name" on-selected-changed="_defaultDataLimitEnabledChanged">
```
fortuna, line 211: For future PRs, it would be better if the renaming to "default data limit" was done in a separate PR.

src/server_manager/web_app/app.ts
```diff
@@ -855,42 +858,82 @@ export class App {
...
     } catch (error) {
-      console.error(`Failed to remove access key data limit: ${error}`);
+      console.error(`Failed to remove server default data limit: ${error}`);
       this.appRoot.showError(this.appRoot.localize('error-remove-data-limit'));
-      serverView.isAccessKeyDataLimitEnabled = true;
+      serverView.isDefaultDataLimitEnabled = true;
+    }
+  }
+
+  private async savePerKeyDataLimit(dialog: OutlinePerKeyDataLimitDialog) {
```
fortuna, line 901: I'm not a big fan of this pattern of passing the dialog back.
Can we instead pass a on-save callback to dialog.open()?

src/server_manager/web_app/app.ts
```diff
@@ -855,42 +858,82 @@ export class App {
...
+      dialog.close();
+      this.appRoot.showNotification(this.appRoot.localize('saved'));
+    } catch (error) {
+      console.error(`Failed to set data limit for access key ${keyId}: ${error}`);
+      dialog.setInitialMenuState();
+      this.appRoot.showError(this.appRoot.localize('error-set-per-key-limit'));
+    }
+  }
+
+  private async removePerKeyDataLimit(dialog: OutlinePerKeyDataLimitDialog) {
```
fortuna, line 921: This could be handled in the same on-save callback above.

fortuna (commented):
Looks good. There are a few minor changes we can make

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
JonathanDCohen: A demonstration of the problem here:

```js
// Easy Case
Number.parseFloat("1.23")
1.23
// First issue
Number.parseFloat("1,23")
1
// Second issue
Number.parseFloat("١٫٢٣")
NaN
```

The first issue we could solve pretty easily. The second issue is a problem I've wrestled with and punted on.  I think it's *possible* to build a localized number parser based on `Intl.NumberFormat.formatToParts`, but that seems like a problem for a different PR, and anyways we haven't heard of any issues with users having trouble with the default data limit input.  We could match the default data limit input and only allow integers, although I'd like in that case to make the same restriction on the server so there isn't skew between the display and the server state.

It would be nice for `Intl` to have localized parsing to go with its localized formatting.

src/server_manager/web_app/ui_components/outline-server-settings.js
JonathanDCohen: This was my mistake in mismanaging my branches.  https://github.com/Jigsaw-Code/outline-server/pull/771 was only merged into a feature base branch I was attempting to use, but never into master.

src/server_manager/web_app/app.ts
```diff
@@ -652,24 +657,26 @@ export class App {
...
+    view.serverVersion = version;
+    view.defaultDataLimitBytes = server.getDefaultDataLimit()?.bytes;
+    view.isDefaultDataLimitEnabled = view.defaultDataLimitBytes !== undefined;
     view.showFeatureMetricsDisclaimer = server.getMetricsEnabled() &&
-        !server.getAccessKeyDataLimit() && !hasSeenFeatureMetricsNotification();
+        !server.getDefaultDataLimit() && !hasSeenFeatureMetricsNotification();
 
-    const version = server.getVersion();
     if (version) {
       view.isAccessKeyPortEditable = semver.gte(version, CHANGE_KEYS_PORT_VERSION);
```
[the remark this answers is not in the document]
JonathanDCohen: Agreed, although this is all sort of glue code anyways.  If we're going to do that maybe we replace most of this method with a `.showServer(server)` method on the view.  That would break our "don't leak server details into the UI" rule though.  It's worth thinking about this.

src/server_manager/web_app/app.ts
JonathanDCohen: Done in be249e6

src/server_manager/web_app/app.ts
JonathanDCohen: Done in be249e6

src/server_manager/web_app/app.ts
```diff
@@ -759,31 +766,28 @@ export class App {
...
+      // Update all the displayed access keys, even if usage didn't change, in case data limits did.
+      const keys = await selectedServer.listAccessKeys();
+      for (const key of keys) {
+        serverView.updateAccessKeyRow(
+            key.id,
+            {transferredBytes: usageMap.get(key.id) ?? 0, dataLimitBytes: key.dataLimit?.bytes});
       }
+      const keyTransferMax = Math.max(0, ...keyTransfers);
+      const defaultDataLimit =
+          serverView.isDefaultDataLimitEnabled ? selectedServer.getDefaultDataLimit() : undefined;
```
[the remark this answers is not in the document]
JonathanDCohen: Good catch! Done in a895e06

src/server_manager/web_app/app.ts
```diff
@@ -759,31 +766,28 @@ export class App {
...
+      // Use a default value for each entry, as any `undefined` will force Math.max to return NaN.
+      const dataLimitMax = Math.max(
```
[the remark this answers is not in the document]
JonathanDCohen: Done in a895e06

src/server_manager/web_app/app.ts
JonathanDCohen: Fixed in 382f7ef

src/server_manager/web_app/app.ts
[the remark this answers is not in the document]
fortuna: My point is to hide how we determine if a feature is available. The `Server` model would have methods like `isAccessKeyPortEditable()`
All those version checks should be hidden. The user of Server shouldn't have to be exposed to that, it's an implementation detail.

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
```diff
@@ -0,0 +1,380 @@
...
+  /**
+   * @member _keyName The displayed name of the UI access key representing the key we're working on.
+   */
+  @internalProperty() _keyName = '';
+  /** @member _keyId The id of the UI access key representing the key we're working on. */
+  @internalProperty() _keyId = '';
+  /**
+   * @member _keyDataLimitBytes The data limit, if it exists, on the access key we're working on.
+   */
+  @internalProperty() _keyDataLimitBytes: number = undefined;
```
fortuna, line 136: Make this a function. It's derived state, no need to maintain this property.

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
```diff
@@ -0,0 +1,380 @@
...
+  /**
+   * @member serverDefaultLimitBytes The default data limit of the server we're working on, or null
+   */
+  @internalProperty() _serverDefaultLimitBytes: number = undefined;
```
fortuna, line 140: Remove this property.

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
```diff
@@ -0,0 +1,380 @@
...
+   * Returns true if the data limit was changed by the user.
+   */
+  public dataLimitChanged() {
+    return this._dataLimitChange() !== Change.UNCHANGED;
+  }
+
+  /**
+   * The current data limit as input by the user, but not necessarily as saved.
+   */
+  public inputDataLimit(): DisplayDataAmount {
```
fortuna, line 318: Get rid of this method and return the data limit value and unit in the `SavePerKeyDataLimitRequested` event instead.

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
```diff
@@ -0,0 +1,380 @@
...
+  /**
+   * Calculates what type of change, or none at all, the user made.
+   */
+  private _dataLimitChange(): Change {
+    if (this._showMenu) {
+      if (this._keyDataLimitBytes === undefined) {
+        // The user must have clicked the checkbox and input a limit.
+        return Change.SET;
+      }
+      const inputLimit = displayDataAmountToBytes(this.inputDataLimit());
```
fortuna, line 295: displayDataAmountToBytes doesn't belong in the UI. Move it to the app code instead.

fortuna (changes requested):
I did a proper review of OutlinePerKeyDataLimitDialog. There are some things to fix there, but they are mostly contained in that file and some code in the App.

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
```diff
@@ -0,0 +1,380 @@
...
+      await this.updateComplete;
+      this._input?.focus();
+    }
+  }
+
+  private _setSaveButtonDisabledState() {
+    this._enableSave = !this._input?.invalid ?? false;
+  }
+
+  private _sendSaveEvent() {
```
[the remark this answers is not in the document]
JonathanDCohen: Obsolete anyways since we don't send an event anymore as of 96bb3353cfdeb80ff5367a78d913f72ed942c1af

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
JonathanDCohen: I find it cleaner to deal with `number` as a vocabulary type in between components instead of passing around `DisplayDataAmount`s.  This way the app doesn't need to know or care how the dialog is formatting the data limit, and the UI and the server don't need to know what the other one is doing. Using bytes in `DisplayAccessKey` as well instead of `DisplayDataAmount` helped us in d5dd4af as well.

  All of the byte <=> display code can live in the UI and then we don't need to worry about particular roundings or display logic in event handlers, `refreshTransferStats`, `convertToUiAccessKey`, etc.

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
JonathanDCohen: See below, I think it's possible to see either the bytes amount or the `DisplayDataAmount` as the derived property and it's cleaner to keep the formatting logic contained in the UI instead of split between the App and the dialog.

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
JonathanDCohen: See below

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
```diff
@@ -0,0 +1,380 @@
...
+    return bytesToDisplayDataAmount(this._activeDataLimit())?.unit || 'GB';
+  }
+
+  private _initialValue() {
+    return bytesToDisplayDataAmount(this._activeDataLimit())?.value.toString() || '';
+  }
+
+  private _activeDataLimit(): number|undefined {
+    // Returns the limit which currently is enforced on this key, or undefined if there is none.
+    return this._keyDataLimitBytes ?? this._serverDefaultLimitBytes;
```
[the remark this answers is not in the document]
JonathanDCohen: Done in 9832382

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
```diff
@@ -0,0 +1,380 @@
...
+          color: #fff;
+        }
+      `,
+    ];
+  }
+
+  /**
+   * @member _keyName The displayed name of the UI access key representing the key we're working on.
+   */
+  @internalProperty() _keyName = '';
```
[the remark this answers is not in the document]
JonathanDCohen: I don't understand how we'd do that in a useful way -- `keyName` isn't known at the time the dialog is constructed and added to the dom, so we'd still need to set it in `open()` anyways.  Alternatively we make the dialog immutable and destroy and re-create new ones instead of opening and closing a singleton, but even in that case we'd have to pass `keyName` to the constructor.

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
```diff
@@ -0,0 +1,380 @@
...
+   * The current data limit as input by the user, but not necessarily as saved.
+   */
+  public inputDataLimit(): DisplayDataAmount {
+    return this._showMenu ? {unit: this._dataLimitType(), value: this._dataLimitValue()} : null;
+  }
+
+  /**
+   * The ID of the key being worked on.  Useful for making API requests for the given key.
+   */
+  public keyId(): string {
```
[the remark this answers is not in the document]
JonathanDCohen: We no longer use this at all

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
```diff
@@ -0,0 +1,380 @@
...
+    return this._keyId;
+  }
+
+  /**
+   * The ID of the server the current key belongs to.
+   */
+  public serverId(): string {
+    return this._serverId;
```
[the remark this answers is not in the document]
JonathanDCohen: We don't use this as of 96bb335

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
```diff
@@ -0,0 +1,380 @@
...
+  @internalProperty() _showMenu = false;
+  /**
+   * @member _enableSave Whether the save button is enabled.  Controlled by the validator on the
+   * input.
+   */
+  @internalProperty() _enableSave = false;
+  /**
+   * @member _language The language used for i18n
+   */
+  @internalProperty() _language = 'en';
```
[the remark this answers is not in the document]
JonathanDCohen: Done in 3c8b91a

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
```diff
@@ -0,0 +1,380 @@
...
+  private async _setCustomLimitTapped() {
+    this._showMenu = !this._showMenu;
+    if (this._showMenu) {
+      await this.updateComplete;
+      this._input?.focus();
+    }
+  }
+
+  private _setSaveButtonDisabledState() {
+    this._enableSave = !this._input?.invalid ?? false;
```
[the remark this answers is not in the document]
JonathanDCohen: `_enableSave` needs to be a property so that we can listen to it and trigger a render if the input is invalid.  Something like 1.2.3 is allowed to by typed by the `allowed-pattern` regex since it checks character-by-character, not the full input string.

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
```diff
@@ -0,0 +1,380 @@
...
+  @internalProperty() _keyDataLimitBytes: number = undefined;
+  /**
+   * @member serverDefaultLimitBytes The default data limit of the server we're working on, or null
+   */
+  @internalProperty() _serverDefaultLimitBytes: number = undefined;
+  /**
+   * @member _showMenu Whether the menu for inputting the data limit should be shown. Controlled by
+   * the checkbox.
+   */
+  @internalProperty() _showMenu = false;
```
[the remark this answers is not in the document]
JonathanDCohen: Done in 7e9c3f3

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
```diff
@@ -0,0 +1,380 @@
...
+    this._language = language;
+    this.setInitialMenuState();
+    this._setSaveButtonDisabledState();
+    this._queryAs<PaperDialogElement>('#container').open();
+  }
+
+  /**
+   * Sets the input state of the dialog back to what it was when it was first opened.
+   */
+  public setInitialMenuState() {
```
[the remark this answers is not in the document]
JonathanDCohen: Done in 96bb3353cfdeb80ff5367a78d913f72ed942c1af

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
```diff
@@ -0,0 +1,380 @@
...
+   *
+   * @param keyName - The displayed name of the access key.
+   * @param keyId - The unique ID of the access key.
+   * @param keyDataLimit - The display data limit of the access key, or undefined.
+   * @param serverId - The ID of the server the access key is a part of.
+   * @param serverDefaultLimit - The default data limit for the server, or undefined if there is
+   *     none
+   */
+  public open(
+      keyName: string, keyId: string, keyLimitBytes: number, serverId: string,
```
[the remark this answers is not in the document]
JonathanDCohen: What I don't like about this idea is that it relies on the caller to satisfy the dialog's internal invariants -- specifically that all of these values are set and **especially** that the callbacks are set whenever the dialog is open.  Using a public method as the single entry point helps us ensure that the dialog is always in a safe state when it's open.

src/server_manager/web_app/app.ts
```diff
@@ -759,31 +766,32 @@ export class App {
...
+      const largerOf = (a: number, b: number|undefined) => {
+        // This check assumes that a is always non-negative, otherwise it fails the b === 0 case.
+        if (!b) {
+          return a;
         }
-        serverView.updateAccessKeyRow(accessKeyId, {transferredBytes, relativeTraffic});
+        return a > b ? a : b;
+      };
+      let keyTransferMax = 0;
+      let dataLimitMax = largerOf(0, selectedServer.getDefaultDataLimit()?.bytes);
```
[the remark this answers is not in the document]
JonathanDCohen: Done in b2ecd5d1a646eaef68037e5280078b0a62b66f49

src/server_manager/web_app/app.ts
```diff
@@ -759,31 +766,32 @@ export class App {
...
+      const keys = await selectedServer.listAccessKeys();
+      for (const key of keys) {
```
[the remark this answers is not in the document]
JonathanDCohen: Done in b2ecd5d1a646eaef68037e5280078b0a62b66f49

src/server_manager/web_app/app.ts
```diff
@@ -759,31 +766,32 @@ export class App {
...
+        serverView.updateAccessKeyRow(
+            key.id,
+            {transferredBytes: usageMap.get(key.id) ?? 0, dataLimitBytes: key.dataLimit?.bytes});
+        keyTransferMax = largerOf(keyTransferMax, usageMap.get(key.id));
+        dataLimitMax = largerOf(dataLimitMax, key.dataLimit?.bytes);
```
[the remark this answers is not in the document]
JonathanDCohen: Done in b2ecd5d1a646eaef68037e5280078b0a62b66f49

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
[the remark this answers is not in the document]
JonathanDCohen: See below

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
```diff
@@ -0,0 +1,380 @@
...
+      `,
+    ];
+  }
+
+  /**
+   * @member _keyName The displayed name of the UI access key representing the key we're working on.
+   */
+  @internalProperty() _keyName = '';
+  /** @member _keyId The id of the UI access key representing the key we're working on. */
+  @internalProperty() _keyId = '';
```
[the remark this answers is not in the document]
JonathanDCohen: See below

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
```diff
@@ -0,0 +1,380 @@
...
+
+  private get _input(): PaperInputElement {
+    return this._queryAs<PaperInputElement>('#dataLimitInput');
+  }
+
+  private _dataLimitValue() {
+    return Number(this._input?.value) ?? 0;
+  }
+
+  private _dataLimitType() {
```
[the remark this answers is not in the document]
JonathanDCohen: Done in 38ea6342ea458b6ef976623f1861eb0a277b8afa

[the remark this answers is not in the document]
JonathanDCohen: Done in 38ea6342ea458b6ef976623f1861eb0a277b8afa

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
```diff
@@ -0,0 +1,380 @@
...
+        </paper-dropdown-menu>
+      </div>
+    `;
+  }
+
+  private _queryAs<T extends HTMLElement>(selector: string): T {
+    return this.shadowRoot.querySelector(selector) as T;
+  }
+
+  private get _input(): PaperInputElement {
```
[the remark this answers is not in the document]
JonathanDCohen: See above.

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
```diff
@@ -197,6 +193,8 @@ export class OutlinePerKeyDataLimitDialog extends LitElement {
   private renderMenu() {
```
fortuna, line 193: Perhaps `renderDataLimit` to align the language?

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
```diff
@@ -269,100 +262,82 @@ export class OutlinePerKeyDataLimitDialog extends LitElement {
...
+      keyName: string, keyLimitBytes: number, onDataLimitSet: OnSetDataLimitHandler,
+      onDataLimitRemoved: OnRemoveDataLimitHandler) {
     this._keyName = keyName;
-    this._keyId = keyId;
-    this._keyDataLimitBytes = keyLimitBytes;
-    this._serverId = serverId;
-    this._serverDefaultLimitBytes = defaultLimitBytes;
-    this._language = language;
-    this.setInitialMenuState();
+    this._activeDataLimitBytes = keyLimitBytes;
```
fortuna, line 325: Let's rename this to `initialDataLimitBytes`. It makes it clearer it was the initial value. "Active" is a higher level concept.

fortuna (changes requested):
Looks good. Just some smaller changes

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
JonathanDCohen: Done in 3a84a661f51d59193d732ae58743c5e6c7075746

src/server_manager/web_app/ui_components/outline-per-key-data-limit-dialog.ts
JonathanDCohen: Done in 3a84a661f51d59193d732ae58743c5e6c7075746

fortuna (approved):
Looks good. Please make sure to merge master, run the app once again and go over the flow to make sure all is good before submitting.

JonathanDCohen:
Thanks for the reminder.  Everything looks good. I've filed a few small UI bugs, mostly which don't have to do with this feature, but I don't think any of them should block rolling this out.

PR #881, opened 2021-04-28
Load per key data limits from the config file

JonathanDCohen:
Fixes a bug where we don't load the data limits on individual keys from the config file when restarting the server.  This would result in dropping the data limits, while instead we want them to be properly persisted.


alalamav (approved):
Good catch!

Issue #888, opened 2021-05-01
Unable to remove existing server in ubuntu


myokyawhtun:
After I downsized the vps server in Google Cloud Platform, the existing VPS server IP got changed where the outline server is running. When I went into Outline manager and the item under "Servers" left sidebar where Outline server from Google Cloud Platform shows non-stop spinning icon. After for few seconds, the "Server unreachable" message shows up and I tried to click "REMOVE SERVER" and it didn't work either. 

![Server unreachable](https://user-images.githubusercontent.com/295096/116792565-77435200-aae7-11eb-9d98-aeb8e2614b50.png)

The Outline version number is 1.8.0 and I use Outline-Manager.AppImage to make it app runnable in Ubuntu. I attached both of the screenshots from Ubuntu.

![developer-tool-outline](https://user-images.githubusercontent.com/295096/116792570-81fde700-aae7-11eb-9496-3be17f387397.png)

Issue #886, opened 2021-04-30
Can't remove unreachable servers from Outline Manager


jli:
Hey y'all :)

My Outline Manager is currently in a weird state:
- There are 2 unreachable servers (GCP, not Digital Ocean) in the Manager.
- Their icons on the left sidebar keep spinning.
- At first, clicking each icon shows the usual "Connections" and "Settings" tabs but with null/0/blank metrics.
  - Clicking the "remove server" menu item from the 3 dot menu doesn't do anything. 
- Eventually, I see this "Server unreachable" page (screenshot attached).
  - Clicking the "Remove Server" button doesn't do anything. 

<img width="810" alt="outline_manaager_bad_state" src="https://user-images.githubusercontent.com/133466/116760851-1da62f00-a9e4-11eb-8c67-9b67d8774193.png">

I got into this state because I was experimenting with starting my own Outline server on GCP. When I restarted the GCE instance, the IP changed, and then this "Server unreachable" issue appeared. I re-ran the install script on the instance, added the new IP to Outline Manager, then again shutdown the server and then had a 2nd unreachable server.

Outline manager version: 1.8.0 (1.8.0.4401)

Thanks!

jli:
Ah, looks like a dupe of https://github.com/Jigsaw-Code/outline-server/issues/873

Issue #873, opened 2021-04-21
Unable to add new server or remove old one


devhops:
I have created a new outline server. That works great.

When I open the Manager to add it, the application hangs. Eventually it times out, but I cannot remove the old server, or add a new one. The application just hangs. I have tried uninstalling but this old instance is never removed.

It would appear others on Reddit are reporting similar issues.

https://www.reddit.com/r/outlinevpn/comments/mqo0g6/i_cannot_remove_the_server_from_outline_manager/ 

For me this is a show stopper as I cannot release it to the client until I can manage it.

I am running MacOS 10.15.7. The Outline Manager version is 1.7.2 (1.7.2.4399).

JonathanDCohen:
Hey there!  Sorry for the delay.

So reading your post, you created the server separately, not from Outline Manager?  Are you able to run the Manager in debug mode and see what comes up in the console?

dylankeyes:
I'm having the same problem trying to remove a server from the Manager that was created in the Manager. This error shows up in the console:

`C:\Users\USER\AppData\Local\Programs\outline-manager\resources\app.asar\node_modules\@sentry\browser\dist\helpers.js:91 Uncaught TypeError: Cannot read property 'forget' of undefined
    at Lc.forgetServer (app.ts:120)
    at HTMLElement.<anonymous> (app.ts:176)
    at HTMLElement.sentryWrapped (C:\Users\USER\AppData\Local\Programs\outline-manager\resources\app.asar\node_modules\@sentry\browser\dist\helpers.js:74)
    at HTMLElement.removeServer (outline-server-view.js:1125)
    at HTMLElement.<anonymous> (template-stamp.js:136)
    at HTMLElement.sentryWrapped (C:\Users\USER\AppData\Local\Programs\outline-manager\resources\app.asar\node_modules\@sentry\browser\dist\helpers.js:74)
    at Qn (gestures.js:669)
    at ei (gestures.js:1065)
    at Object.click (gestures.js:1024)
    at HTMLElement.Kn (gestures.js:457)
    at HTMLElement.sentryWrapped (C:\Users\USER\AppData\Local\Programs\outline-manager\resources\app.asar\node_modules\@sentry\browser\dist\helpers.js:74)`

JonathanDCohen:
That's no good.

[...]

PR #890, opened 2021-05-04
Populate an unreachable server's view with the server id

JonathanDCohen:
This fixes issues forgetting an unreachable server which were caused by the missing id keeping us from finding the server to forget.

Fix https://github.com/Jigsaw-Code/outline-server/issues/888
Fix https://github.com/Jigsaw-Code/outline-server/issues/886
Fix https://github.com/Jigsaw-Code/outline-server/issues/873


alalamav (approved):
👍

PR #910, opened 2021-06-03
Improve GCP "billing account" screen

bemasc:
This change makes it clearer that user action is required, and polls for
changes in billing status in order to save 1-2 clicks.


bemasc:
Before:
Screenshot from 2021-06-03 16-22-02

After:
Screenshot from 2021-06-03 16-21-03

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
```diff
@@ -98,6 +98,7 @@ export class GcpCreateServerApp extends LitElement {
   private project: Project;
   private billingAccounts: BillingAccount[] = [];
   private regionPicker: OutlineRegionPicker;
+  private billingAccountsRefreshLoop: NodeJS.Timeout = null;
```
JonathanDCohen, line 101: This is web code, not Node code.  You can use Number for this.  If tsc complains using `window.setTimeout` should fix the problem.

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
```diff
@@ -207,21 +208,23 @@ export class GcpCreateServerApp extends LitElement {
...
     return html`
       <outline-step-view id="billingAccountSetup" display-action="">
-        <span slot="step-title">Activate your Google Cloud Platform account.</span>
-        <span slot="step-description">Enter your billing information on Google Cloud Platform.</span>
+        <span slot="step-title">Add billing information to your Google Cloud Platform account.</span>
+        <span slot="step-description">Open the Cloud Console billing page and add an account in order to proceed.</span>
         <span slot="step-action">
-          <paper-button id="createServerButton" @tap="${this.handleBillingVerificationNextTap}">
-            NEXT
+          <paper-button id="openBillingPage" @tap="${this.openBillingPage}">
```
JonathanDCohen, line 214: Maybe a way to go about this is to automatically start the polling, show "OPEN" when the polling is happening, and "NEXT" so the user can move on when the polling stops.  Otherwise there will be a gap where the user is stuck waiting for the next poll.

JonathanDCohen (changes requested):

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
bemasc: That was the purpose of the REFRESH button.  I changed the name to "CHECK AGAIN" to try to make that clearer:

Screenshot from 2021-06-04 12-17-02

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
bemasc: Thanks, done.

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
```diff
@@ -207,21 +208,23 @@ export class GcpCreateServerApp extends LitElement {
   private renderBillingAccountSetup() {
     return html`
       <outline-step-view id="billingAccountSetup" display-action="">
-        <span slot="step-title">Activate your Google Cloud Platform account.</span>
-        <span slot="step-description">Enter your billing information on Google Cloud Platform.</span>
+        <span slot="step-title">Add billing information to your Google Cloud Platform account.</span>
```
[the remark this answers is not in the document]
bemasc: I think that's true for this whole flow.

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
```diff
@@ -207,21 +208,23 @@ export class GcpCreateServerApp extends LitElement {
...
+        <span slot="step-description">Open the Cloud Console billing page and add an account in order to proceed.</span>
         <span slot="step-action">
-          <paper-button id="createServerButton" @tap="${this.handleBillingVerificationNextTap}">
-            NEXT
+          <paper-button id="openBillingPage" @tap="${this.openBillingPage}">
+            OPEN
```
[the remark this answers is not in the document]
bemasc: Done!

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
```diff
@@ -303,32 +307,57 @@ export class GcpCreateServerApp extends LitElement {
...
+      window.bringToFront();
+    }
   }
 
-  private async showProjectSetup(existingProject?: Project): Promise<void> {
-    this.billingAccounts = await this.account.listBillingAccounts();
-    if (!this.billingAccounts || this.billingAccounts.length === 0) {
-      return this.showBillingAccountSetup();
+  private stopRefreshingBillingAccounts(): void {
+    if (this.billingAccountsRefreshLoop !== null) {
```
JonathanDCohen, line 350: nit: you can `clearInterval(null)` and it will just do nothing.

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
```diff
@@ -303,32 +307,57 @@ export class GcpCreateServerApp extends LitElement {
...
 
   private showBillingAccountSetup(): void {
     this.currentPage = 'billingAccountSetup';
   }
 
-  private async handleBillingVerificationNextTap(): Promise<void> {
-    this.showProjectSetup();
+  private async refreshBillingAccounts(): Promise<void> {
+    this.billingAccounts = await this.account.listBillingAccounts();
+    // FIXME: listBillingAccounts() can reject, resulting in an uncaught
```
JonathanDCohen, line 338: If we want to put in some UI for now we can send an event to be handled by appRoot and use `showError`, something like https://github.com/Jigsaw-Code/outline-server/blob/master/src/server_manager/web_app/app.ts#L272.

JonathanDCohen (commented):
LGTM with a little nit.  Thanks for the update!

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
bemasc: I adjusted this comment.  I think a toast is a reasonable possibility, but I'd prefer to land this PR and discuss the error handling question separately.

JonathanDCohen (approved):

src/server_manager/messages/master_messages.json
```diff
@@ -443,6 +443,26 @@
...
+  "gcp_billing_action": {
+    "message": "OPEN",
+    "description": "Button label for a button that opens a the billing page in the web browser"
+  },
+  "gcp_billing_body": {
+    "message": "Waiting for you to add a billing account on Google Cloud",
+    "description": "Body text on the billing page"
+  },
+  "gcp_billing_refresh": {
+    "message": "CHECK AGAIN",
```
fortuna, line 463: Should this follow proper case? We can do upper case with the CSS. What do we use on the other buttons? Let's be consistent.

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
```diff
@@ -303,32 +307,55 @@ export class GcpCreateServerApp extends LitElement {
     } else {
```
fortuna, line 307: For readability, change to `else if` to avoid the extra nesting.

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
```diff
@@ -303,32 +307,55 @@ export class GcpCreateServerApp extends LitElement {
...
       if (!this.billingAccounts || this.billingAccounts.length === 0) {
         this.showBillingAccountSetup();
+        // Check every five seconds to see if an account has been added.
+        this.billingAccountsRefreshLoop = window.setInterval(() => {
```
fortuna, line 311: You need to stop a previous loop if there was one running.

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
```diff
@@ -303,32 +307,55 @@ export class GcpCreateServerApp extends LitElement {
...
+          this.refreshBillingAccounts();
+        }, 5000);
       } else {
         this.showProjectSetup(this.project);
       }
     }
   }
 
+  disconnectedCallback() {
```
fortuna, line 320: Does this component actually get disconnected? Some components are just hidden. It depends on the parent

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
```diff
@@ -303,32 +307,55 @@ export class GcpCreateServerApp extends LitElement {
...
     }
+  }
+
+  private stopRefreshingBillingAccounts(): void {
+    window.clearInterval(this.billingAccountsRefreshLoop);
+    this.billingAccountsRefreshLoop = null;
+  }
 
+  private openBillingPage(): void {
+    window.open("https://console.cloud.google.com/billing");
```
fortuna, line 355: How do we open urls in other places of the code?

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
```diff
@@ -303,32 +307,55 @@ export class GcpCreateServerApp extends LitElement {
...
+
+    if (this.billingAccounts && this.billingAccounts.length > 0) {
+      this.stopRefreshingBillingAccounts();
+      this.showProjectSetup();
+      window.bringToFront();
     }
+  }
+
+  private stopRefreshingBillingAccounts(): void {
+    window.clearInterval(this.billingAccountsRefreshLoop);
```
fortuna, line 350: Do you need window?

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
```diff
@@ -303,32 +307,55 @@ export class GcpCreateServerApp extends LitElement {
...
     this.currentPage = 'billingAccountSetup';
   }
 
-  private async handleBillingVerificationNextTap(): Promise<void> {
-    this.showProjectSetup();
-  }
-
-  private async showProjectSetup(existingProject?: Project): Promise<void> {
+  private async refreshBillingAccounts(): Promise<void> {
     this.billingAccounts = await this.account.listBillingAccounts();
```
fortuna, line 337: Please catch and log the error. We shouldn't have uncaught exceptions.

Failing silently is ok, since the Next click will check and tell the user the result of the check.

fortuna (commented):
Thanks for tackling this. We need to make some changes. I'm sorry for the delay.

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
bemasc: Currently it does not, so this code is just for future-proofing or contractual correctness.

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
bemasc: It is stopped by `this.init()`.

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
bemasc: OK, I added catch-and-log blocks to both places where this is called.

I don't think silent failures are OK, but we should think more about what the right solution is.  If the call is persistently failing (as happened to me once, maybe due to an expired oauth token), the user will be stuck at this page, without any indication that something is wrong.

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
bemasc: No.  We need `window` for the call to `setInterval` due to the TypeScript bindings, so I used `window` here for consistency.

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
bemasc: This is similar to how it's done in browser_main.ts.  I changed this one to make them more similar (and fixed an error in the other one).

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
bemasc: Done

src/server_manager/messages/master_messages.json
bemasc: You're right.  Changed to match the other button labels.

bemasc (commented):
OK, changes are in #915.

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
fortuna: I agree. I made a suggestion to address that in the new PR.

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
fortuna: Isn't this a leak?
Doesn't that mean that you will keep polling forever?
And keep adding new loops every time you open the create server screen?

fortuna: Perhaps we can check the visibility in the loop somehow. I don't remember how we do in the other parts of the app.

bemasc: Reopening the create server screen will call `start()` on the existing element.  `start()` stops any existing polling loop via `this.init()`.  The loop also stops once you add an account.

It's true that the loop will continue if you hide the GCP "app" and never add a billing account.

fortuna: Here is what you can do:
- Add a stop() method
- In app-root.js, add a handler for currentPage changes that will call the stop method.

This is exactly what we do in the outline-server-view to manage the animation:
https://github.com/Jigsaw-Code/outline-server/blob/e86c1981ed9ec8ce4d96be2bd088b73b8dbfad48/src/server_manager/web_app/ui_components/outline-server-view.js#L973

It's possible to set a on-changed handler on the iron-page, or a property observer on currentPage.

PR #911, opened 2021-06-03
Fix GCP billing account list visibility problem

bemasc:
When billing accounts are present with long names, they would be cut off
on the left side.  This shifts the list to extend right instead of left.


bemasc:
Before:
Screenshot from 2021-06-03 17-14-08

After:
Screenshot from 2021-06-03 17-12-25

fortuna (approved):

PR #912, opened 2021-06-03
Prevent duplicate server creation in GCP

bemasc:
This change disables the "create server" button in GCP when server
creation is in progress.


bemasc:
Before:
Screenshot from 2021-06-04 10-43-43

After:
Screenshot from 2021-06-04 10-44-47

fortuna (approved):

PR #923, opened 2021-06-15
Update to electron-13

bemasc:
This change updates the electron version, and other dependency versions
that needed to be updated due to compatibility or deprecation.


fortuna (approved):
Thanks for the upgade!

PR #934, opened 2021-06-26
Update install_server.sh

Qtong121:
in BSD tr dont have the option --delete but have -d


Qtong121:
@googlebot I signed it!

fortuna (approved):
Thanks for the update

PR #938, opened 2021-07-08
Add flags for all countries

bemasc:
This fills in the missing country flags for GCP.  It replaces the custom
flag images with images from the circle-flags library.


bemasc:
Before:
Screenshot from 2021-07-08 15-35-06

After:
Screenshot from 2021-07-08 13-40-45

fortuna (changes requested):
Let's add the border and the gradient. Here's the spec for the CSS:
image

Don't use the CSS as is. We can simplify it  bit. I don't think we need the transform or change the positioning, for example.

src/server_manager/web_app/ui_components/outline-region-picker-step.ts
```diff
@@ -25,17 +25,7 @@ import {COMMON_STYLES} from './cloud-install-styles';
...
-  'SG': `${FLAG_IMAGE_DIR}/singapore.png`,
-  'UK': `${FLAG_IMAGE_DIR}/uk.png`,
-  'DE': `${FLAG_IMAGE_DIR}/germany.png`,
-  'NL': `${FLAG_IMAGE_DIR}/netherlands.png`,
-  'CA': `${FLAG_IMAGE_DIR}/canada.png`,
-  'US': `${FLAG_IMAGE_DIR}/us.png`,
-};
 
 @customElement('outline-region-picker-step')
 export class OutlineRegionPicker extends LitElement {
```
fortuna, line 31: Can you add a class to the subtitle and make it use the secondary text color and upper case?
This will make it closer to the mocks.

fortuna (commented):
Thanks for this! It looks a lot better without all the unknowns. I think we can tweak the UI to be closer to the mocks.

fortuna:
CSS in text to make it easier:
```css
background: linear-gradient(180deg, rgba(20, 20, 20, 0.2) 0%, rgba(0, 0, 0, 0) 100%);
mix-blend-mode: normal;
border: 4px solid rgba(255, 255, 255, 0.1);
```

src/server_manager/web_app/ui_components/outline-region-picker-step.ts
bemasc: Done

bemasc:
It now looks like this:
Screenshot from 2021-07-08 21-03-32

fortuna (approved):

PR #943, opened 2021-07-20
Fix settings page for GCP

bemasc:
This change removes the DigitalOcean name and logos from GCP servers.


src/server_manager/web_app/ui_components/outline-gcp-info-setting.ts
```diff
@@ -0,0 +1,53 @@
...
+import {customElement, html, LitElement, property} from 'lit-element';
+
+import {COMMON_STYLES} from './cloud-install-styles';
+import {CloudLocation} from '../../model/location';
+import {getShortName} from '../location_formatting';
+import {SETTINGS_STYLES} from './outline-server-settings-styles';
+
+/** Setting card for displaying GCP-specific info */
+@customElement('outline-gcp-info-setting')
+export class OutlineGcpInfoSetting extends LitElement {
```
fortuna, line 28: I don't understand. Why do you need this element? Can't you reuse the DO one and pass the name and icon as input?

src/server_manager/web_app/ui_components/outline-gcp-info-setting.ts
```diff
@@ -0,0 +1,53 @@
...
+           See https://github.com/lit/lit-element/issues/633. -->
+      <style>${SETTINGS_STYLES}</style>
+      <div class="setting card-section">
+        <img class="setting-icon" src="images/gcp-logo.svg">
+        <div>
+          <h3>Google Cloud</h3>
+          <paper-input readonly always-float-label maxlength="100"
+              .value="${getShortName(this.cloudLocation, this.localize)}"
+              .label="${this.localize('settings-server-location')}"
+              .hidden="${!this.cloudLocation}"></paper-input>
```
fortuna, line 49: Omit the html instead of using hidden.

src/server_manager/web_app/ui_components/outline-server-settings.js
```diff
@@ -159,29 +77,17 @@ Polymer({
...
-        <div class="setting card-section" hidden\$="[[!isServerManaged]]">
-          <img class="setting-icon digital-ocean-icon" src="images/do_white_logo.svg">
-          <div>
-            <h3>DigitalOcean</h3>
-            <paper-input readonly="" value="[[_getShortName(cloudLocation, localize)]]" label="[[localize('settings-server-location')]]" hidden\$="[[!cloudLocation]]" always-float-label="" maxlength="100"></paper-input>
-            <paper-input readonly="" value="[[serverMonthlyCost]]" label="[[localize('settings-server-cost')]]" hidden\$="[[!serverMonthlyCost]]" always-float-label="" maxlength="100"></paper-input>
-            <paper-input readonly="" value="[[serverMonthlyTransferLimit]]" label="[[localize('settings-transfer-limit')]]" hidden\$="[[!serverMonthlyTransferLimit]]" always-float-label="" maxlength="100"></paper-input>
-          </div>
-        </div>
+        <outline-do-info-setting hidden\$="[[!_isDo(cloudId)]]"
```
fortuna, line 83: As mentioned in the other file, there's no need for the separate elements. You can pass the cloud name and icon.

src/server_manager/web_app/ui_components/outline-server-settings.js
```diff
@@ -362,7 +268,13 @@ Polymer({
     return valid ? '' : this.localize('error-keys-port-bad-input');
   },
 
-  _getShortName: getShortName,
+  _isDo(cloudId) {
+    return cloudId === DO_CLOUD_ID;
```
fortuna, line 272: No need for cloud id. Pass the cloud name and icon.

Perhaps we need a cloud_assets library with methods like getDigitaloceanIcon()

src/server_manager/web_app/app.ts
```diff
@@ -755,6 +756,7 @@ export class App {
       view.monthlyOutboundTransferBytes =
           host.getMonthlyOutboundTransferLimit()?.terabytes * (10 ** 12);
       view.cloudLocation = host.getCloudLocation();
+      view.cloudId = host.getCloudId();
```
fortuna, line 759: Set the cloud name and icon instead, based on the Server type.

src/server_manager/model/digitalocean.ts
```diff
@@ -15,6 +15,8 @@
 import * as location from "./location";
 import {ManagedServer} from "./server";
 
+export const DO_CLOUD_ID = 'do';
```
fortuna, line 18: There's no need for the model to have cloud ids. We have per-cloud code in app.ts. Every caller of addServer is cloud-specific. Whatever needs to be cloud-specific, can live in that code. For example you can define cloud ids in the app layer and pass that in addServer. addServer can map that id to cloud name and icon.

We have a view per server, created at addServer time. We can leverage that and set all the parameters we need there.

On a side note, the way we do the server views are suboptimal. Instead of using data binding, we are adding views explicitly. If we had a ui model and used binding it would be a bit cleaner, but that's a big change.

fortuna (changes requested):

src/server_manager/model/digitalocean.ts
fortuna: If you think it will make things easier, perhaps we can have methods like `isDigitaloceanServer(server)` possibly based on some `server.isDigitalOcean()`, since interfaces don't exist in javascript (you'd check the presence of the method).

src/server_manager/web_app/ui_components/outline-server-settings-styles.js
```diff
@@ -34,36 +35,135 @@ styleElement.appendChild(html`
...
     --paper-input-container-underline-disabled: {
       display: none;
-    }
+    };
     --paper-input-container-disabled: {
       opacity: 1;
-    }
+    };
+  }
+  paper-input:not([readonly]) {
```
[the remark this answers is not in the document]
bemasc: The DigitalOcean views are unchanged.  For GCP, it looks like this
Screenshot from 2021-07-30 13-51-27
Screenshot from 2021-07-30 13-51-32

src/server_manager/web_app/ui_components/outline-server-view.js
```diff
@@ -489,12 +490,12 @@ export class ServerView extends DirMixin(PolymerElement) {
               </div>
               <p>[[localize('server-data-transfer')]]</p>
             </div>
-            <div hidden\$="[[!isServerManaged]]" class="stats-card card-section">
+            <div hidden\$="[[!_showUtilizationPercentage(cloudId, monthlyOutboundTransferBytes)]]" class="stats-card card-section">
```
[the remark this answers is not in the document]
bemasc: Done

src/server_manager/web_app/ui_components/outline-do-info-setting.ts
```diff
@@ -0,0 +1,63 @@
...
+           See https://github.com/lit/lit-element/issues/633. -->
+      <style>${SETTINGS_STYLES}</style>
+      <div class="setting card-section">
+        <img class="setting-icon" src="images/do_white_logo.svg">
+        <div>
+          <h3>DigitalOcean</h3>
+          <paper-input readonly always-float-label maxlength="100"
+              .value="${getShortName(this.cloudLocation, this.localize)}"
+              .label="${this.localize('settings-server-location')}"
+              .hidden="${!this.cloudLocation}"></paper-input>
```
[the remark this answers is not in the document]
bemasc: I've removed this file.

src/server_manager/web_app/ui_components/outline-gcp-info-setting.ts
bemasc: OK, done.

src/server_manager/web_app/ui_components/outline-gcp-info-setting.ts
bemasc: I've removed this file.

src/server_manager/web_app/ui_components/outline-server-settings.js
bemasc: Done

src/server_manager/web_app/ui_components/outline-server-settings.js
bemasc: OK, added a cloud_assets library with `getCloudIcon(DisplayCloudId)`.

src/server_manager/web_app/app.ts
bemasc: I've moved the cloud ID to the account, which better maps it association.

src/server_manager/web_app/ui_components/cloud-assets.ts
```diff
@@ -0,0 +1,42 @@
...
+    case DisplayCloudId.DO:
+      return 'images/do_white_logo.svg';
+    case DisplayCloudId.GCP:
+      return 'images/gcp-logo.svg';
+    default:
+      return null;
+  }
+}
+
+export function getCloudName(id: DisplayCloudId): string {
```
fortuna, line 33: Could you add a TODO to internationalize?

src/server_manager/web_app/ui_components/outline-server-view.js
```diff
@@ -489,12 +490,12 @@ export class ServerView extends DirMixin(PolymerElement) {
               </div>
               <p>[[localize('server-data-transfer')]]</p>
             </div>
-            <div hidden\$="[[!isServerManaged]]" class="stats-card card-section">
+            <div hidden\$="[[!monthlyOutboundTransferBytes]]" class="stats-card card-section">
               <div>
-                <img class="digital-ocean-icon" src="images/do_white_logo.svg">
+                <img class="digital-ocean-icon" src="[[getCloudIcon(cloudId)]]">
```
fortuna, line 495: Please rename the class name.

src/server_manager/web_app/ui_components/cloud-assets.ts
```diff
@@ -0,0 +1,42 @@
...
+  DO = 'do',
+  GCP = 'gcp'
+}
+
+export function getCloudIcon(id: DisplayCloudId): string {
+  switch (id) {
+    case DisplayCloudId.DO:
+      return 'images/do_white_logo.svg';
+    case DisplayCloudId.GCP:
+      return 'images/gcp-logo.svg';
```
fortuna, line 27: No need to fix it in your PR, but it would be nice to have a consistent file name for these files.

src/server_manager/web_app/app.ts
```diff
@@ -334,7 +336,8 @@ export class App {
       this.digitalOceanAccount = digitalOceanAccount;
       this.appRoot.digitalOceanAccount = {
         id: this.digitalOceanAccount.getId(),
-        name: await this.digitalOceanAccount.getName()
+        name: await this.digitalOceanAccount.getName(),
+        cloudId: DisplayCloudId.DO,
```
fortuna, line 340: The DO account object lives in a DO-specific variable. A cloudId doesn't belong there, since its cloud is clear.

Use DisplayCloudId directly in app-root's `_getCloudId`

src/server_manager/web_app/ui_components/app-root.js
```diff
@@ -1009,6 +1011,32 @@ export class AppRoot extends mixinBehaviors
...
+   */
+  _getCloudId(accountId) {
+    // TODO: Replace separate account fields with a map.
+    if (this.gcpAccount && accountId === this.gcpAccount.id) {
+      return this.gcpAccount.cloudId;
+    } else if (this.digitalOceanAccount && accountId === this.digitalOceanAccount.id) {
+      return this.digitalOceanAccount.cloudId;
+    }
+    return null;
+  }
```
fortuna, line 1026:
```suggestion
  _getCloudId(accountId) {
    // TODO: Replace separate account fields with a map.
    if (this.gcpAccount && accountId === this.gcpAccount.id) {
      return DisplayCloudId.GCP;
    } else if (this.digitalOceanAccount && accountId === this.digitalOceanAccount.id) {
      return DisplayCloudId.DO;
    }
    return null;
  }
```

src/server_manager/web_app/gallery_app/main.ts
```diff
@@ -114,12 +126,12 @@ export class TestApp extends LitElement {
     this.setLanguage('en');
   }
 
-  async setLanguage(newLanguage: string) {
+  setLanguage(newLanguage: string) {
```
fortuna, line 129: Why are you changing this?

fortuna (changes requested):

src/server_manager/web_app/ui_components/cloud-assets.ts
bemasc: Google and DO don't translate these names, so I don't think we can or should.  Until we add a cloud that translates its own name, I don't think we need a TODO.

src/server_manager/web_app/ui_components/outline-server-view.js
bemasc: Done

src/server_manager/web_app/ui_components/cloud-assets.ts
bemasc: Also one of them is actually white, and the other is color that is being forced to grayscale with CSS.

src/server_manager/web_app/app.ts
bemasc: Done.  (Note that we'll need something like this for multi-account support.)

src/server_manager/web_app/gallery_app/main.ts
bemasc: `outline-server-view` depends on having `this.language` correct at startup.  The asynchronous `setLanguage` call meant that `this.language` was `''` when `outline-server-view` loads, resulting in errors and failure to load.

src/server_manager/web_app/app.ts
fortuna: It depends on the multi-cloud implementation. For instance, you can still have a per-cloud account list

src/server_manager/web_app/gallery_app/main.ts
fortuna: Could you add a comment. It's not clear.

Also, this seems like a bug in the outline-server-view. All components should be able to have their language changed.

An alternative is to place the `<outline-server-view>` inside an if-block that tests if the language and localize are present.

fortuna (approved):

src/server_manager/web_app/gallery_app/main.ts
bemasc: The main issue is that `outline-server-view` makes heavy use of `this.localize` and `this.language`, resulting in uncaught exceptions because these are not initialized yet.  (Polymer somehow tolerates this when they are used in a template.)

I've reverted this change, and instead made the relevant Polymer elements tolerant of null/empty `localize` and `language` properties.  I've also added various missing bindings so that changes are tracked correctly.

src/server_manager/web_app/ui_components/outline-server-view.js
```diff
@@ -854,12 +856,21 @@ export class ServerView extends DirMixin(PolymerElement) {
     }));
   }
 
+  // Avoids uncaught exceptions when this.localize is transiently null,
+  // as is the case when using the widget gallery app.
+  _localize() {
```
fortuna, line 861: You are introducing a new pattern to localize the strings, which can confuse the reader. It will make the reader wonder when to use this and when to use localize directly, which we do everywhere else.

Isn't it simpler and clearer to set a dummy localize function in the constructor?

src/server_manager/web_app/data_formatting.ts
```diff
@@ -49,6 +49,9 @@ function getDataFormattingParams(numBytes: number): FormatParams {
 function makeDataAmountFormatter(language: string, params: FormatParams) {
   // We need to cast through `unknown` since `tsc` mistakenly omits the 'unit' field in
   // `NumberFormatOptions`.
+  if (!language) {
```
fortuna, line 52: Let's revert this. Language is required and it should be the caller's responsibility to enforce that.

src/server_manager/web_app/data_formatting.ts
```diff
@@ -75,7 +78,11 @@ export function formatBytesParts(numBytes: number, language: string): DataAmount
     throw new Error('formatBytesParts only works in web app code. Node usage isn\'t supported.');
   }
   const params = getDataFormattingParams(numBytes);
-  const parts = makeDataAmountFormatter(language, params).formatToParts(params.value);
+  const formatter = makeDataAmountFormatter(language, params);
+  if (!formatter) {
```
fortuna, line 82: Same here. We should just assume the language is present, otherwise we will be hiding a bug. In the callers that need you can check the language before calling this (as we already do in outline-server-view)

src/server_manager/web_app/ui_components/outline-server-settings.js
```diff
@@ -359,16 +361,24 @@ Polymer({
...
+    }
+    return '';
   },
 
   _getShortName: getShortName,
+  _getCloudIcon: getCloudIcon,
+  _getCloudName: getCloudName,
 
   _getInternationalizedUnit(bytesAmount, language) {
     return formatBytesParts(bytesAmount, language).unit;
```
fortuna, line 375: Perhaps check the language here.

src/server_manager/web_app/gallery_app/main.ts
fortuna: I don't get why this wasn't a problem before. Isn't there a default localize function set? If not, can't we set a dummy localize in the app root constructor?

I see you are passing the localize function in some template calls that depend on it, which is a good fix.

For the unit stuff, check the presence of the language in the caller instead (like we already do in some places), so we don't hide any bugs. It's up to the caller to decide what to do in those cases.

fortuna (changes requested):

src/server_manager/web_app/ui_components/outline-server-view.js
bemasc: I had this, but reverted it along with the change to `setLanguage`.  I've reinstated the dummy localizer approach.  To make this work without changes to `setLanguage`, we also need the initial value of `language` to be valid, but not `'en'`, which would cause `setLanguage('en')` to short-circuit.  I've set it to `es-419` as a placeholder.

src/server_manager/web_app/data_formatting.ts
bemasc: Reverted

src/server_manager/web_app/data_formatting.ts
bemasc: Reverted

src/server_manager/web_app/ui_components/outline-server-settings.js
bemasc: I would prefer to demand that `language` is always valid.

src/server_manager/web_app/gallery_app/main.ts
```diff
@@ -84,15 +93,15 @@ const GCP_BILLING_ACCOUNTS: gcp.BillingAccount[] =
 @customElement('outline-test-app')
 export class TestApp extends LitElement {
   @property({type: String}) dir = 'ltr';
-  @property({type: Function}) localize: (...args: string[]) => string;
+  @property({type: Function}) localize: (...args: string[]) => string = fakeLocalize;
+  @property({type: String}) language = 'es-419';  // Replaced asynchronously in the constructor.
```
fortuna, line 97: Perhaps use something invalid like `zz`?

fortuna (approved):
Looks great. Thanks for the changes!

src/server_manager/web_app/gallery_app/main.ts
bemasc: Done.  (Curious that `''` throws an exception but `'zz'` is allowed.)

PR #944, opened 2021-07-20
Improve UI for accessing GCP "beta" mode

bemasc:
This PR:
* Adds a promo link for GCP "beta" mode to the GCP manual instructions
* Removes a step by auto-advancing to the correct state after GCP login (matching our DO login behavior)
* Adds text to the GCP intro box in the logged-in state
* Replaces text in the GCP intro box when GCP_AUTH_ENABLED=true


bemasc:
With this PR, the card looks like this:

Initial state:
Screenshot from 2021-07-20 16-30-50

Logged in:
Screenshot from 2021-07-20 16-30-33

fortuna (changes requested):
Could you make this controlled by a flag or variable so we can easily switch?
I would be good to allow people to try it before we actually release it to everyone, since it may be broken in some ways.

bemasc:
Good point.  I've updated this PR so the UI is controlled by the GCP_AUTH_ENABLED flag.

bemasc:
I've also added a promo link to the GCP instructions page:
Screenshot from 2021-07-21 14-27-36

src/server_manager/messages/master_messages.json
```diff
@@ -1099,6 +1103,10 @@
     "message": "Set up Outline anywhere",
     "description": "This string appears in the server setup view as a header within a card. The card provides information about installing Outline on an arbitrary cloud server provider."
   },
+  "setup_beta": {
+    "message": "Beta",
```
fortuna, line 1107: Drop "Beta" and use "Experimental" instead, so we are consistent

src/server_manager/web_app/app.ts
```diff
@@ -602,6 +602,7 @@ export class App {
     } else {
       await this.showDigitalOceanCreateServer(this.digitalOceanAccount);
     }
+    bringToFront();
```
fortuna, line 605: Why is this?

src/server_manager/web_app/app.ts
```diff
@@ -620,8 +621,13 @@ export class App {
       return;
     }
 
-    await this.loadGcpAccount(gcpAccount);
-    this.showIntro();
+    const gcpServers = await this.loadGcpAccount(gcpAccount);
+    if (gcpServers.length > 1) {
```
fortuna, line 625: I think you meant this:
```suggestion
    if (gcpServers.length > 0) {
```

src/server_manager/web_app/app.ts
```diff
@@ -620,8 +621,13 @@ export class App {
...
+      this.showServer(gcpServers[0]);
+    } else {
+      this.appRoot.getAndShowGcpCreateServerApp().start(this.gcpAccount);
+    }
+    bringToFront();
```
fortuna, line 630: This can be removed, I believe. The user already has the screen in the foreground.

src/server_manager/web_app/ui_components/outline-intro-step.js
```diff
@@ -205,15 +205,39 @@ Polymer({
           </div>
         </div>
 
-        <div id="gcp" class="card" on-tap="setUpGcpTapped">
+        <div id="gcp" class="card" on-tap="setUpGcpTapped" hidden\$="[[!_showNewGcpFlow(gcpAccountName)]]">
           <div class="card-header">
-            <div class="tag" hidden\$="[[_computeIsAccountConnected(gcpAccountName)]]">[[localize('setup-advanced')]]</div>
+            <div class="tag">[[localize('setup-beta')]]</div>
```
fortuna, line 210: Use "Experimental" instead.

src/server_manager/web_app/ui_components/outline-manual-server-entry.js
```diff
@@ -188,6 +202,10 @@ Polymer({
               <iron-icon id="gcpCreateServerDropDownIcon" icon="arrow-drop-down"></iron-icon>
             </div>
           </div>
+          <div id="gcp-new-flow-promo" class="section-content-instructions" on-tap="gcpNewFlowTapped">
```
fortuna, line 205: Please move this to above the first card. It shouldn't look as part of the manual flow, but a escape hatch.

src/server_manager/web_app/ui_components/outline-manual-server-entry.js
```diff
@@ -127,6 +127,20 @@ Polymer({
           display: none;
         }
       }
+      #gcp-tag {
+        font-weight: 500;
+        letter-spacing: 0.05em;
+        font-size: 10px;
+        margin-bottom: 8px;
+        text-transform: uppercase;
+        color: #4285f4;
```
fortuna, line 136: Don't introduce a new color. Use the same color we use in the titles.

src/server_manager/web_app/ui_components/outline-manual-server-entry.js
```diff
@@ -188,6 +202,10 @@ Polymer({
               <iron-icon id="gcpCreateServerDropDownIcon" icon="arrow-drop-down"></iron-icon>
             </div>
           </div>
+          <div id="gcp-new-flow-promo" class="section-content-instructions" on-tap="gcpNewFlowTapped">
+            <div id="gcp-tag">[[localize('experimental')]]</div>
+            <a>[[localize('setup-gcp-promo')]]<iron-icon icon=open-in-new></iron-icon></a>
```
fortuna, line 207: The "open in new window" icon is not appropriate, since we open in the same window. We should remove it.

Instead, convert the link to text and add a button with the set up server call to action (same we use for DO)

src/server_manager/web_app/ui_components/outline-intro-step.js
```diff
@@ -290,6 +313,10 @@ Polymer({
     return Boolean(accountName);
   },
 
+  _showNewGcpFlow(gcpAccountName) {
+    return outline.gcpAuthEnabled || this._computeIsAccountConnected(gcpAccountName);
```
fortuna, line 317: Let's remove the `outline.gcpAuthEnabled` check. If an account is connected, we should always show the new flow.

This check should be in the manual screen instead, to control when to show the experimental flow escape hatch.

src/server_manager/web_app/ui_components/outline-intro-step.js
```diff
@@ -205,15 +205,39 @@ Polymer({
...
           <div class="card-header">
-            <div class="tag" hidden\$="[[_computeIsAccountConnected(gcpAccountName)]]">[[localize('setup-advanced')]]</div>
+            <div class="tag">[[localize('setup-beta')]]</div>
             <div class="email" hidden\$="[[!_computeIsAccountConnected(gcpAccountName)]]">[[gcpAccountName]]</div>
             <img src="images/gcp-logo.svg">
           </div>
           <div class="card-title">Google Cloud Platform</div>
           <div class="card-body">
-            <div class="description" hidden\$="[[_computeIsAccountConnected(gcpAccountName)]]">
+          <div class="description">
```
fortuna, line 216: This is a bit hard to read. Perhaps move the card htmls to functions?

fortuna (changes requested):

src/server_manager/messages/master_messages.json
bemasc: Done

src/server_manager/web_app/app.ts
bemasc: This is returning from the OAuth flow, which leaves the user looking at a blank tab in their main browser.

src/server_manager/web_app/app.ts
bemasc: I don't think that's true.  The user is looking at the blank redirect target at the end of the OAuth flow, as a result of calling `this.runGcpOauthFlow()`.

src/server_manager/web_app/ui_components/outline-intro-step.js
bemasc: Done

src/server_manager/web_app/ui_components/outline-intro-step.js
bemasc: > If an account is connected, we should always show the new flow.

We do.  (This is an `||`.)  The purpose of checking `gcpAuthEnabled` is to show the magic-flow promo card in the logged-out state.

> This check should be in the manual screen instead, to control when to show the experimental flow escape hatch.

The manual screen for GCP is unreachable when `gcpAuthEnabled` is true, so that seems moot.

src/server_manager/web_app/ui_components/outline-manual-server-entry.js
bemasc: Done.  (That color was --gcp-blue, but the variable isn't available in this scope.)

src/server_manager/web_app/ui_components/outline-manual-server-entry.js
bemasc: Done

src/server_manager/web_app/ui_components/outline-manual-server-entry.js
bemasc: Clicking this link instantly opens the Google Oauth flow in a new window, so the icon seems natural to me.

src/server_manager/web_app/app.ts
fortuna: Interesting. I think we used to bring to front in the past.
In that case, let's bringToFront right after OAuth.

src/server_manager/web_app/app.ts
fortuna: Let's move this to right after the OAuth.

src/server_manager/messages/master_messages.json
fortuna: I still see the message here. Delete?

src/server_manager/web_app/ui_components/outline-intro-step.js
fortuna: I still see setup-beta

src/server_manager/web_app/ui_components/outline-manual-server-entry.js
fortuna: I still see the blue

fortuna (changes requested):
I don't think you pushed all the changes. I still see unmodified files.

src/server_manager/web_app/app.ts
bemasc: Done

src/server_manager/web_app/app.ts
bemasc: Done

src/server_manager/web_app/ui_components/outline-intro-step.js
bemasc: OK, I moved the card HTML into constants.  (Templates are static in polymer.)  This required fixing the postcss hook to handle multiple `<style>` tags in one file.

src/server_manager/web_app/ui_components/outline-intro-step.js
fortuna: You're right, I misread it.

fortuna (approved):

PR #945, opened 2021-07-20
Mark GCP Free Tier regions as "Lowest Cost"

bemasc:


bemasc:
Looks like this:
Screenshot from 2021-07-20 17-28-12

fortuna:
@cjhenck thoughts on the look?

Perhaps we can use a [corner ribbon](https://codepen.io/eode9/pen/twkKm), but the current implementation serves the purpose.

src/server_manager/web_app/digitalocean_account.ts
```diff
@@ -60,7 +60,8 @@ export class DigitalOceanAccount implements digitalocean.Account {
     const regions = await this.digitalOcean.getRegionInfo();
     return regions.map(info => ({
       cloudLocation: new digitalocean.Region(info.slug),
-      available: info.available && info.sizes.indexOf(MACHINE_SIZE) !== -1
+      available: info.available && info.sizes.indexOf(MACHINE_SIZE) !== -1,
+      lowerCost: false,  // All DO regions have the same prices.
```
fortuna, line 64: This is the wrong semantics. They should all have this set to true. You can revert this file.

src/server_manager/web_app/gcp_account.ts
```diff
@@ -84,10 +92,14 @@ export class GcpAccount implements gcp.Account {
...
-    return zones.map(zoneInfo => ({
-      cloudLocation: new gcp.Zone(zoneInfo.name),
-      available: zoneInfo.status === 'UP'
-    }));
+    return zones.map(zoneInfo => {
+      const zone = new gcp.Zone(zoneInfo.name);
+      return {
+        cloudLocation: zone,
+        available: zoneInfo.status === 'UP',
+        lowerCost: FREE_TIER_REGIONS.has(zone.regionId),
```
fortuna, line 100: You need to update the `gcp.ZoneInfo` type.

You will want to update `filterOptions` to be a generic that takes <T extends CloudLocationOption> and returns T[] in order to not lose the typing.

In `GcpCreateServerApp.showRegionPicker()` you will want to convert to the UI input type.

fortuna (changes requested):
I'm not a big fan of changing the shared model because of a GCP-specific need. But I have some suggestions there.

src/server_manager/model/location.ts
```diff
@@ -87,4 +87,5 @@ export interface CloudLocation {
 export interface CloudLocationOption {
   readonly cloudLocation: CloudLocation;
   readonly available: boolean;
+  readonly lowerCost: boolean;
```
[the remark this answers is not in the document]
bemasc: Removed.  We don't really have a `ZoneInfo` type.  I moved it to `RegionPickerOption` instead.

src/server_manager/web_app/digitalocean_account.ts
bemasc: Reverted

src/server_manager/web_app/gcp_account.ts
bemasc: `zoneInfo` is a type defined by GCP.  Instead of returning the attribute here, I made it a check that GcpCreateServerApp can perform, and made `filterOptions` generic as suggested.

src/server_manager/web_app/ui_components/outline-region-picker-step.ts
```diff
@@ -144,6 +153,7 @@ export class OutlineRegionPicker extends LitElement {
           <input type="radio" id="card-${index}" name="city" value="${index}" ?disabled="${!option.available}" .checked="${this.selectedIndex === index}" @change="${this._locationSelected}">
           <label for="card-${index}" class="city-button">
             <div class="card-header">
+              <div class="tag">${option.lowerCost ? this.localize('region-lowest-cost') : ''}</div>
```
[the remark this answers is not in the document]
bemasc: It's not possible to remove that dependency, because `GeoLocation` is used throughout the stack.  However, I have removed this field from `CloudLocationOption`, and defined it in a new `RegionPickerOption` instead.

src/server_manager/web_app/ui_components/outline-region-picker-step.ts
```diff
@@ -27,9 +27,13 @@ import {getShortName, localizeCountry} from '../location_formatting';
 
 const FLAG_IMAGE_DIR = 'images/flags';
 
+export interface RegionPickerOption extends CloudLocationOption {
```
fortuna, line 30: Please add a TODO to remove the dependency on the model.
Or define this type independently. The caller should convert from the domain model to the view model.

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts
```diff
@@ -364,7 +364,12 @@ export class GcpCreateServerApp extends LitElement {
     // `this.regionPicker` is null after `this.currentPage`, and is only populated
     // asynchronously.
     this.regionPicker = this.shadowRoot.querySelector('#regionPicker') as OutlineRegionPicker;
-    this.regionPicker.options = filterOptions(zoneOptions);
+    const displayOptions = filterOptions(zoneOptions);
```
fortuna, line 367: This is not yet RegionPickerOption, so let's not call it `displayOptions`.

Also, let's not modify the return value of listLocations. That could potentially affect future calls. You need to read that function to make sure things are not breaking. Instead, make a copy as part of the conversion to RegionPickerOption.

fortuna (changes requested):
Oops, this fell through the cracks. Thanks for the change. One minor thing left.

src/server_manager/web_app/ui_components/outline-region-picker-step.ts
bemasc: I added a TODO.

Even if we remove the dependency on `CloudLocationOption`, I think we still want to share the `CloudLocation` type across the whole codebase, so I don't think it's as simple as "remove the dependency".  However, I also think this code may change: for GCP, we really should treat this as a "location picker", and let the server creation flow choose the zone _after_ the location (i.e. GCP Region) is selected.  That would help with a problem I've encountered several times, where the selected zone is "available" but actually has no capacity when we try to create the instance.

src/server_manager/web_app/ui_components/outline-gcp-create-server-app.ts

fortuna (approved):

PR #958, opened 2021-08-11
Install Docker on GCP using the Ubuntu dpkg method

bemasc:
This cuts the GCP server setup time approximately in half, likely
because it avoids multiple calls to "apt update", which
synchronizes all known package repositories.

This PR also updates the underlying Ubuntu image to the latest LTS
release.


fortuna (approved):
Excellent!

PR #959, opened 2021-08-11
Consolidate duplicated "Disconnect" strings

bemasc:
Our translation system requires extra work to support distinct string
with identical content, and in this case there's no need for distinct
strings anyway.


fortuna (approved):
That's surprising to me. Don't we use the message ids?

bemasc:
Our translation pipeline appears to ignore these message ids, and requires a separate field ("meaning") to indicate that identical strings are actually distinct.

PR #961, opened 2021-08-12
Switch instance type to e2-micro

bemasc:
e2-micro is the new GCP Free Tier instance type.  It's also faster and
has more RAM than the old f1-micro type.


fortuna (approved):
Is it available on all regions?

bemasc:
Yes, it's available everywhere.

PR #967, opened 2021-08-19
Improve GCP instance name construction

bemasc:
The current code generates names by concatenating integers in a custom
format, with several bugs:
* Leading zeros are discarded
* The date is in local time, but the time is in UTC.
* Months are presented zero-indexed (i.e. January is month zero)

This change corrects these issues.


fortuna (approved):
I think that `20210112-130150` would be a bit more readable than `2021-01-12t13-01-50`.
Or perhaps we just replace the `t` with a dash: `2021-01-12-13-01-50`

bemasc:
OK, I've restored the previous format but with bugfixes, and updated the PR description.

fortuna (approved):

PR #982, opened 2021-09-09
Fix login bug after free trial period expires

bemasc:
When the free trial period expires, the Outline project is left linked
to an inactive billing account, and all instances are deleted.
Attempting to list the instances associated with a project that lacks
billing results in an HTTP 403.  Currently, this results in GCP login
failure, with a generic error message.

This change makes the manager handle this account state gracefully,
taking the user to the billing setup screen, and from there directly
to the server creation screen.

### DIFF
--- a/src/server_manager/web_app/app.ts
+++ b/src/server_manager/web_app/app.ts
@@ -376,7 +376,7 @@ export class App {
           result.push(server);
         }
       } catch (e) {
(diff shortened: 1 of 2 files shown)


bemasc:
@JonathanDCohen FYI

fortuna (approved):

fortuna (approved):

PR #996, opened 2021-10-06
Allow custom name for shadow docker

ksperis:
This allow to launch multiple instances on same host.

### DIFF
--- a/src/server_manager/install_scripts/install_server.sh
+++ b/src/server_manager/install_scripts/install_server.sh
@@ -19,6 +19,9 @@
 
 # You may set the following environment variables, overriding their defaults:
 # SB_IMAGE: The Outline Server Docker image to install, e.g. quay.io/outline/shadowbox:nightly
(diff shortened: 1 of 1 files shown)


ksperis:
@fortuna The use case is for multiple teams. We have one manager for each team who is independent to manage the keys assigned to the users of his team.

fortuna:
@daniellacosse can you test the install script from this fork and make sure it works in a clean machine?

Instruction at https://github.com/Jigsaw-Code/outline-server/tree/master/src/shadowbox#self-hosted-installation, but use their fork instead.

I'm quite busy this week and won't be able to test it until late tomorrow.

daniellacosse:
> @daniellacosse can you test the install script from this fork and make sure it works in a clean machine?
> 
> Instruction at https://github.com/Jigsaw-Code/outline-server/tree/master/src/shadowbox#self-hosted-installation, but use their fork instead.
> 
> I'm quite busy this week and won't be able to test it until late tomorrow.
Screen Shot 2021-10-19 at 3 04 07 PM

PR #998, opened 2021-10-07
[ATTEMPTING NODE 14 UPGRADE OVER CI]

daniellacosse:
Mainly it seems the critical changes were made to the Streams API, which we don't depend on


daniellacosse:
@fortuna seems I need a linux box to repro this error - https://github.com/Jigsaw-Code/outline-server/pull/998/checks?check_run_id=3831899108#step:10:254

works fine on MacOS:
Screen Shot 2021-10-07 at 4 40 30 PM

any intuition about what might be going awry here? @bemasc

fortuna:
I wonder if this is the problem: https://stackoverflow.com/a/67755255/1234111

PR #1116, opened 2022-06-27
chore(manager): show the GCP magic flow by default

bemasc:
"Advanced" GCP flow (now legacy) is available by setting
GCP_AUTH_ENABLED=false.


daniellacosse (approved):
So I understand, there's no way to turn a flag off in production, we'd have to build again?

PR #1130, opened 2022-07-28
fix(manager): remove dev tools

daniellacosse:
https://git.radicallyopensecurity.com/ros/google-outline/-/issues/7

Seems to do the trick, don't see the dev tools here on mac anyway:

Screen Shot 2022-07-28 at 14 48 27


fortuna (approved):
Thanks for another quick fix!!

PR #1132, opened 2022-08-02
fix: avoid malicious path traversal in manager's open-image method

kevindamm-jigsaw:
There is an Electron IPC event 'open-image' that accepts any path,
including relative paths that would expose system filenames such as
/etc/passwd.  This change resolves all paths to be rooted at the base
directory for ...server_manager/web_app.


fortuna (approved):
Thanks for the fix!

PR #1138, opened 2022-08-08
fix(manager): removes special case for manager connection key

daniellacosse:
Screen Shot 2022-08-08 at 12 01 30

should resolve https://git.radicallyopensecurity.com/ros/google-outline/-/issues/21


daniellacosse:
What should I test to confirm this doesn't break anything?

src/server_manager/web_app/ui_components/outline-server-view.ts
```diff
@@ -908,14 +825,6 @@ export class ServerView extends DirMixin(PolymerElement) {
     }
   }
 
-  // Help bubbles should be shown after this outline-server-view
-  // is on the screen (e.g. selected in iron-pages). If help bubbles
-  // are initialized before this point, setPosition will not work and
-  // they will appear in the top left of the view.
-  showGetConnectedHelpBubble() {
```
fortuna, line 915: Humm, we added this for a reason. I think it was to address the issue that managers weren't sure what to do next. I think it would be better to keep the bubble.

daniellacosse: Should I add into the regular access key row, so it'll be there for each instance?

fortuna: I guess so, but it should only ever show on key 0, if it wasn't dismissed before.

fortuna (commented):
For testing, you can make sue the key 0 is working. Also try adding a few keys, then delete key 0 to make sure nothing weird happens.

daniellacosse:
> I guess so, but it should only ever show on key 0, if it wasn't dismissed before.

Here's it now:

Screen Shot 2022-08-15 at 16 56 56

fortuna (approved):

PR #1157, opened 2022-08-26
fix: don't close the oauth callback server on bad requests

kevindamm-jigsaw:
Changes the behavior of the local server for receiving oauth callbacks
from Digital Ocean by not closing it on invalid requests.  This avoids
the problem where a malicious actor prematurely closes the
callback-handling server (a DoS of the authentication flow).

Also added a timeout to close the server after 30 seconds.

### DIFF
--- a/src/server_manager/electron_app/digitalocean_oauth.ts
+++ b/src/server_manager/electron_app/digitalocean_oauth.ts
@@ -151,8 +151,6 @@ export function runOauth(): OauthSession {
     // This is the POST endpoint that receives the access token and redirects to either DigitalOcean
     // for the user to complete their account creation, or to a page that closes the window.
     app.post('/', express.urlencoded({type: '*/*', extended: false}), (request, response) => {
-      server.close();
-
       const params = new URLSearchParams(request.body.params);
       if (params.get('error')) {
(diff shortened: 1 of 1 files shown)


fortuna (dismissed):
It seems there's an error in the current code.

@daniellacosse @jyyi1  can you do the code review?

src/server_manager/electron_app/digitalocean_oauth.ts
```diff
@@ -209,6 +208,9 @@ export function runOauth(): OauthSession {
         }
         reject(error);
       });
+
+    // Automatically close the server after 30 seconds.
+    setTimeout(server.close, 30000);
```
daniellacosse, line 213: Will this send an error/do we want to send a 408? Also, mind moving the 30000 to a const at the head of the file or something, I'm iffy on magic nos

kevindamm-jigsaw: This is inside the POST handler for the oauth response, so the error response would go to whoever sent the improper request (or we would return 200 success for the request + correct token from digital ocean).

But that brings up a good point, I should actually move this timeout to right after the server creation, instead of in the POST handler.  I'll also add a const definition for the timeout.

daniellacosse (approved):

PR #1158, opened 2022-08-31
chore: updated translations for Outline manager

kevindamm-jigsaw:
chore: updated translations for Outline manager


src/server_manager/messages/messages.json
```diff
@@ -0,0 +1,1304 @@
+{
```
jyyi1, line 1: We don't need to include this file.

src/server_manager/messages/iw.json
```diff
@@ -0,0 +1,272 @@
+{
```
jyyi1, line 1: Please don't include `iw` as well, it is identical to `he` which is already included. Here is [more information](https://source.corp.google.com/piper///depot/google3/jigsaw/firehook/outline/BUILD;l=25).

kevindamm-jigsaw: done

PR #1166, opened 2022-09-27
Revert "refactor(manager): replace share s3 flow with self-contained instructions"

daniellacosse:
Reverts Jigsaw-Code/outline-server#1133


jyyi1 (approved):
LGTM

kevindamm-jigsaw:
LGTM does this need translations to be reverted too?

daniellacosse:
> LGTM does this need translations to be reverted too?

Ah, yes - It should be a recent PR. Double check that that's a clean revert, too.

kevindamm-jigsaw:
yeah, I think it was #1158
https://github.com/Jigsaw-Code/outline-server/pull/1158

there was another translation change for the price update but I think that doesn't need to be reverted (and was before the above PR so shouldn't conflict, either)

daniellacosse:
> yeah, I think it was #1158
> 
> there was another translation change for the price update but I think that doesn't need to be reverted (and was before the above PR so shouldn't conflict, either)

Awesome, make it so

kevindamm-jigsaw (approved):

PR #1181, opened 2022-10-05
chore: update package.json to set version as 1.12.0

daniellacosse:
to be merged once we update the links in the invitation flow @jyyi1

### DIFF
--- a/src/server_manager/package.json
+++ b/src/server_manager/package.json
@@ -1,7 +1,7 @@
 {
   "name": "outline-manager",
   "productName": "Outline Manager",
-  "version": "1.11.0",
+  "version": "1.12.0",
   "description": "Create and manage access to Outline servers",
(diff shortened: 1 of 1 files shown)


src/server_manager/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "outline-manager",
   "productName": "Outline Manager",
-  "version": "1.11.0",
+  "version": "1.12.0",
```
jyyi1, line 4: Is `1.11.1` better since we did not introduce any new features?

daniellacosse: The last change was a minor bump and we're reverting that, so my logic here was that the revert of a minor patch is also a minor patch.

kevindamm-jigsaw (approved):

PR #1207, opened 2022-10-21
feat(server): include icanhazip as an alternative to ipinfo and checkip of domains.google.com

aaomidi:
Add icanhazip.com as an alternative. Both ipinfo and checkip are sometimes inaccessible from certain areas.


fortuna (approved):
Thanks for your contribution! It seems like Cloudflare runs icanhazip.com now, so we could use it: https://blog.apnic.net/2021/06/17/how-a-small-free-ip-tool-survived/

PR #1237, opened 2022-11-28
fix(manager/web): 💊 show Sydney flag in DO region page

jyyi1:
The Sydney flag and its city name was not displayed appropriately in DigitalOcean's region selection page. It was a simple fix by just adding it to the available regions. Also [`substr` function is deprecated](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr), I replaced it with the [standard `substring` function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring).

**after fix**

image


fortuna (approved):
Thanks for the quick fix!

PR #1273, opened 2023-01-24
feat(server): add optional name/limit/password props for createNewAccessKey method

murka:


fortuna:
@murka Could you add a bit of background on why you need this change?
What are you trying to accomplish?

murka:
@fortuna, I would like to add options when creating a key, right now you need send three steps for create the key, change name and add limits.

murka:
@fortuna This PR is ready for merge!

https://github.com/murka/outline-server/actions/runs/5685731607

vatrubin:
@murka Could you add a password field? I would like to synchronize passwords between several servers, but this cannot be done through the API.

murka:
@vatrubin, no, it's not completion for this PR, password will be automatically generated for each accessKey.

murka:
@fortuna @daniellacosse, any updates?

murka:
Added the password prop in the `AccessKeyCreateParams` interface, see https://github.com/Jigsaw-Code/outline-server/pull/1273/commits/22554db3d453192429bd05df45a22e905fecdee2

src/shadowbox/server/manager_service.ts
```diff
@@ -192,6 +192,24 @@ function validateDataLimit(limit: unknown): DataLimit {
   return limit as DataLimit;
 }
 
+function validatePassword(password: string): string {
+  if (!password) {
```
fortuna, line 196: You're validating the password twice. This is dead code you can remove.

src/shadowbox/server/manager_service.ts
```diff
@@ -192,6 +192,24 @@ function validateDataLimit(limit: unknown): DataLimit {
...
+    throw new restifyErrors.MissingParameterError(
+      {statusCode: 400},
+      'Missing `password` parameter'
+    );
+  }
+
+  const bytesOfPassword = Buffer.byteLength(password as string, 'utf8');
```
fortuna, line 203: Why the `as string`? `password` is already a string.
```suggestion
  const bytesOfPassword = Buffer.byteLength(password, 'utf8');
```

src/shadowbox/model/access_key.ts
```diff
@@ -48,9 +48,20 @@ export interface AccessKey {
   readonly dataLimit?: DataLimit;
 }
 
+export interface AccessKeyCreateParams {
+  // The encryption method to use for the access key.
```
fortuna, line 52: Can you add the id as well?
That will allow people to export/import keys, or create keys for a user on demand, in a way that preserve the id used in the Prometheus metrics. That's a popular request from server operators.

In the implementation, make sure the id doesn't already exist.

src/shadowbox/server/manager_service.ts
```diff
@@ -192,6 +192,24 @@ function validateDataLimit(limit: unknown): DataLimit {
...
+  if (bytesOfPassword < 22) {
```
fortuna, line 204: This kind of business logic should be in the repository instead, in the model layer (in terms of Domain-Driven Design). This file is just an application-layer adaptor to expose the domain objects, it should only do serialization/deserialization and type validation.

I'm also ok with removing the byte length check. If the service provider is using this option, they must know what they are doing, and the limit may get in the way.

src/shadowbox/server/manager_service.ts
```diff
@@ -311,28 +329,61 @@ export class ShadowsocksManagerService {
...
+            `Expected a string name, instead got ${name} of type ${typeof name}`
+          )
+        );
+      }
+
+      const dataLimit = (req.params.limit as DataLimit) || undefined;
+
+      if (dataLimit) {
+        validateDataLimit(dataLimit);
+      }
```
fortuna, line 363: You can't use `as DataLimit` because it may be invalid.

Use the validation function to get the types object:
```suggestion
    const dataLimit DataLimit | undefined
    if (re.params.limit) {
      dataLimit = validateDataLimit(req.params.limit);
    }
```

src/shadowbox/server/manager_service.ts
```diff
@@ -311,28 +329,61 @@ export class ShadowsocksManagerService {
...
+
+      const password = (req.params.password as string) || undefined;
+      if (password) {
+        validatePassword(password);
```
fortuna, line 367: I'd rather return an error, and `return next(error)` here, so the error handling and control flow is clearer.

src/shadowbox/server/api.yml
```diff
@@ -150,7 +150,7 @@ paths:
                   type: string
```
fortuna, line 150: You need to list the new properties here.

src/shadowbox/server/server_access_key.ts
```diff
@@ -182,7 +188,9 @@ export class ServerAccessKeyRepository implements AccessKeyRepository {
       encryptionMethod,
       password,
     };
-    const accessKey = new ServerAccessKey(id, '', metricsId, proxyParams);
+    const name = params?.name ?? this.NEW_USER_DEFAULT_NAME;
+    const dataLimit = params?.dataLimit ?? this.NEW_USER_DEFAULT_DATA_LIMIT;
```
fortuna, line 192: You can remove this line, since the ServerAccessKey has the dataLimit as optional.

Also, the code reads as if we set a value other than `undefined`, which we don't.

src/shadowbox/server/server_access_key.ts
```diff
@@ -182,7 +188,9 @@ export class ServerAccessKeyRepository implements AccessKeyRepository {
       encryptionMethod,
       password,
     };
-    const accessKey = new ServerAccessKey(id, '', metricsId, proxyParams);
+    const name = params?.name ?? this.NEW_USER_DEFAULT_NAME;
```
fortuna, line 191: This is more readable:

```suggestion
    const name = params?.name ?? '';
```

fortuna (changes requested):
Thanks for the changes. Here are a few more comments.

src/shadowbox/model/access_key.ts
fortuna: We can leave the id out of this PR.

src/shadowbox/server/manager_service.ts
fortuna: I still see this here. Let's just remove the length check.
We only need to check if it's a non empty string.

src/shadowbox/server/manager_service.ts
```diff
@@ -311,28 +329,58 @@ export class ShadowsocksManagerService {
...
+        return next(
+          new restifyErrors.InvalidArgumentError(
+            {statusCode: 400},
+            `Expected a string name, instead got ${name} of type ${typeof name}`
+          )
+        );
+      }
+
+      const dataLimit = req.params.limit ? validateDataLimit(req.params.limit) : undefined;
+      const password = req.params.password ? validatePassword(req.params.password) : undefined;
```
fortuna, line 360: This is clearer. It should fail with an empty string password.
```suggestion
      const dataLimit = req.params.limit ?? validateDataLimit(req.params.limit);
      const password = req.params.password ?? validatePassword(req.params.password);
```

fortuna (changes requested):

src/shadowbox/server/manager_service.ts
```diff
@@ -311,28 +329,58 @@ export class ShadowsocksManagerService {
...
+            `Expected a string name, instead got ${name} of type ${typeof name}`
+          )
+        );
+      }
+
+      const dataLimit = req.params.limit ?? validateDataLimit(req.params.limit);
+      const password = req.params.password ?? validatePassword(req.params.password);
+
+      const accessKeyJson = accessKeyToApiJson(
+        await this.accessKeys.createNewAccessKey({
```
fortuna, line 363: We are getting compilation errors in the CI:
image

Run `npm run action shadowbox/server/build` to reproduce.

Please fix it.

src/shadowbox/server/manager_service.ts
```diff
@@ -311,28 +329,58 @@ export class ShadowsocksManagerService {
...
+        return next(
+          new restifyErrors.InvalidArgumentError(
+            {statusCode: 400},
+            `Expected a string name, instead got ${name} of type ${typeof name}`
+          )
+        );
+      }
+
+      const dataLimit = req.params.limit ?? validateDataLimit(req.params.limit);
+      const password = req.params.password ?? validatePassword(req.params.password);
```
fortuna, line 360: I think this needs to be like this to get the type check right:
```suggestion
      const dataLimit = validateDataLimit(req.params.limit);
      const password = validatePassword(req.params.password);
```

fortuna (changes requested):

src/shadowbox/server/manager_service.ts
fortuna: The check for 22 bytes is still there. Please remove. We only need to check if it's a non empty string.

fortuna (commented):
Thanks for fixing the test. I realize that one issue was not addressed.

fortuna:
Oh, there are still tests failing when you run `npm run action shadowbox/test`.
See https://github.com/Jigsaw-Code/outline-server/actions/runs/7469550377/job/20371104279?pr=1273

murka:
@fortuna, all jobs is succeed: https://github.com/murka/outline-server/actions/runs/7369417871

fortuna (approved):
Thanks a lot for the contribution and patience!

src/shadowbox/server/api.yml
```diff
@@ -146,11 +146,19 @@ paths:
             schema:
               type: object
               properties:
+                name:
+                  type: string
                 method:
                   type: string
+                password:
+                  type: string
+                port:
```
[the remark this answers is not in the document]
fortuna: @jadolg I think you are right. You can still set the port for new access keys if you want to do that, but we should add it to the key creation API as well.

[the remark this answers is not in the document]
sbruens: I'll send a PR to add this.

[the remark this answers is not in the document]
jadolg: Please also review the `PUT` method. I think it's wired to the same method so it should not work either but the docs claim it does.

PR #1293, opened 2023-02-18
Add server manager install script to support rootless docker

Friend-LGA:
Right now default install path is using this script
```bash
sudo bash -c "$(wget -qO- https://raw.githubusercontent.com/Jigsaw-Code/outline-server/master/src/server_manager/install_scripts/install_server.sh)"
```

This one doesn't work with **Rootless Docker** because of 2 reasons:
- `--net host` flag
- usage of `watchtower`

This PR adds new script `install_server_rootless.sh` which maps docker ports directly using `-p` flag and removes usage of `watchtower`.

Works with Rootless Docker (tested).

Example usage
```bash
bash -c "SHADOWBOX_DIR=$HOME/outline-data $(wget -qO- https://raw.githubusercontent.com/Jigsaw-Code/outline-server/master/src/server_manager/install_scripts/install_server_rootless.sh)"
```

Notice no `sudo` is here and custom shadowbox dir.

### DIFF
--- a/src/server_manager/install_scripts/install_server_rootless.sh
+++ b/src/server_manager/install_scripts/install_server_rootless.sh
@@ -1,0 +1,538 @@
+#!/bin/bash
+#
+# Copyright 2018 The Outline Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


fortuna (changes requested):
Thanks for the contribution. However, there are a few issues:

- It doesn't fully work. If one changes the port for access keys, Docker won't be mapping those new ports, so it won't work. That's why we use `--net host`. We also need that to determine what ports are in use.
- It loses the history, so I can't tell what the differences are from the original script
- It introduces a lot of duplicate code to maintain

You need to address those issues in order for us to consider merging the contributions.
Why do you need Rootless mode in the first place?

PR #1298, opened 2023-02-27
fix(manager): support for Alpine Linux

Jonathunky:
A couple of fixes: option to use rc-status instead of systemctl, and changed names of logfiles to properly support BusyBox's logging.


fortuna (commented):
Thanks for your contribution. Would you mind making that small adjustment?

fortuna (approved):

PR #1327, opened 2023-04-26
fix(manager): MacOS: fixed link from iTunes to App Store

vvmspace:


vvmspace:
Accept or not accept, all done from my side.

Thank you!

PR #1423, opened 2023-10-11
chore: upgrade restify to v11 to support lts/hydrogen

daniellacosse:


fortuna (approved):
I think it's ok if the integration test pass. We should be extra careful with the server release though, and test using the Manager to make sure there's no issue with CORS.

PR #1451, opened 2023-11-16
fix(manager/electron): separate main and renderer sentry libraries

daniellacosse:
This was causing the JS in the renderer process to crash

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,510 +31,8 @@
         "typescript": "^4"
       },
       "engines": {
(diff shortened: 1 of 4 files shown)


fortuna (approved):
Wow, nice catch. Thanks so much for fixing it!

jyyi1:
We need to let our QA test the feedback form: just let them submit a feedback, and then go to Sentry to see whether it appears. IIRC previously it was not working if you are using the separate library. Maybe they have fixed this issue.

PR #1461, opened 2023-12-11
fix(sentry_webhook): rollback Salesforce value changes

sbruens:
Rolling back #1419. This is a likely culprit of Salesforce not creating cases.


sbruens:
> You're gonna have to set the node engines version to `18.x.x` - see my email entitled "It seems the required node version shifted in the Github Action without warning." Otherwise looks good.

Done.

sbruens:
I think I have a fix for whatever is failing in #1462. I have no idea why it only started failing today. Maybe some cache of OmahaProxy has finally expired?

PR #1466, opened 2024-01-01
fix(docs): fix invalid json in api payload examples

sqeezelemon:


fortuna (approved):
Thanks for the contribution!

PR #1478, opened 2024-01-19
feat(server): add a new state property to a key and expose it in the API

sbruens:
No change in behavior. This allows us to add new states in future changes.

### DIFF
--- a/src/shadowbox/model/access_key.ts
+++ b/src/shadowbox/model/access_key.ts
@@ -15,6 +15,16 @@
 export type AccessKeyId = string;
 export type AccessKeyMetricsId = string;
 
+// The state of an access key, indicating the place in its life cycle.
+export enum AccessKeyState {
+  // The default value if the state is missing or omitted.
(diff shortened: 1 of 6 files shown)


fortuna (changes requested):
This is good as a proof of concept. But until we have a concrete use, I don't think this should be submitted. It's dead code that we need to maintain as part of our API, and we may have to change it when we actually implement what we need.

I'm not convinced with the state modelling. It conflates multiple states in one. For instance, perhaps we want a ENABLED/DISABLED state, with disabled having a reason. QUOTA EXCEEDED/NOT EXCEEDED can be its own separate state.
We probably need a design doc for the end-to-end feature. We've talked about at least [3 different approaches](https://docs.google.com/document/d/1TalEtlsINS3zXrTKV074JJ5euZ5FkCdcZVA6-xgr6mU/edit?resourcekey=0-wYQc4WqJPK5-U_TNQiYzfw&tab=t.28hw0dhvp2vf).

Issue #1499, opened 2024-02-05
Puth method in /access-keys needs to use {id} in the path


jadolg:
The method to create a key with a specific id operates under `/access-keys/{id}` and not under just `/access-keys`
https://github.com/Jigsaw-Code/outline-server/blob/64c4f945342ea7725b936a538f9b39353ee55379/src/shadowbox/server/api.yml#L178

The code is correct but the API docs are not

sbruens:
Thank you @murka for fixing this!

PR #1503, opened 2024-02-06
fix(server): api specification

murka:
fix: #1499


murka:
hm, git diff is not correctly, I move `PUT` method from `access-keys` into `access-keys/{id}`, as per https://github.com/Jigsaw-Code/outline-server/blob/64c4f945342ea7725b936a538f9b39353ee55379/src/shadowbox/server/manager_service.ts#L142

sbruens (approved):
Thank you for fixing this!

sbruens:
Sorry, do you mind resolving the conflicts? Then I'll merge it.

PR #1508, opened 2024-02-13
fix(manager/electron): adds hardened runtime and notarization options to the MacOS build

daniellacosse:
context provided here: https://github.com/Jigsaw-Code/outline-server/pull/1508#issuecomment-1942000793

### DIFF
--- a/src/server_manager/electron_app/release/macos.entitlements
+++ b/src/server_manager/electron_app/release/macos.entitlements
@@ -4,6 +4,8 @@
   <dict>
     <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
     <true/>
+    <key>com.apple.security.cs.allow-jit</key>
+    <true/>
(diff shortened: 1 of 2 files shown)


src/server_manager/electron_app/release/macos.entitlements
```diff
@@ -4,6 +4,8 @@
   <dict>
     <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
     <true/>
+    <key>com.apple.security.cs.allow-jit</key>
```
sbruens, line 7: Just curious. Why was this needed?

sbruens (approved):
Could you add some more details to the PR to explain the "why" of this change? What changed since our last successfully notarized build that these changes were necessary? Would be helpful to have that record for future reference.

sbruens (approved):

PR #1555, opened 2024-06-05
Fix isOverDataLimit logic to stop allowing 0 traffic byte

cornzzy:
[issue #1515 ](https://github.com/Jigsaw-Code/outline-server/issues/1515)

This if statement is just wrong and doesn't block access when traffic limit is set to 0 byte. This one character fixes it.

Now if we create a new access key, set the limit to 0 before first usage, the client no longer will be able to connect.


cornzzy:
Sorry for the duplicate, deleted my fork.

PR #1566, opened 2024-07-18
test(server): fix integration test by downgrading to Alpine 3.18

sbruens:
Alpine 3.19 curl is using the c-ares resolver instead of the system resolver. This is causing [DNS issues in the integration test](https://github.com/Jigsaw-Code/outline-server/actions/runs/9993257424/job/27620375345):

```
curl: (6) Could not resolve host: shadowbox
```

See https://github.com/curl/curl/issues/12558 and https://github.com/c-ares/c-ares/issues/683 for details.

### DIFF
--- a/.github/workflows/build_and_test_debug.yml
+++ b/.github/workflows/build_and_test_debug.yml
@@ -14,10 +14,9 @@ on:
       - master
 
 jobs:
-
-  shadowbox:
-    name: Shadowbox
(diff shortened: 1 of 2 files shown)


fortuna (approved):
Wow, nasty. Good catch. You may want to add a comment to the Dockerfile explaining the issue.

PR #1511, opened 2024-02-21
chore: delete manager and update server_manager/README.md

daniellacosse:

PR #1569, opened 2024-07-25
clean(devtools/build): remove unused jsign files

sbruens:
These were part of the server manager build, which was moved out of this repo in #1511 and Jigsaw-Code/outline-apps#1834.


jyyi1 (approved):
Thanks for the clean up! We don't need jsign now since the Manager has been merged to outline-apps.

PR #1581, opened 2024-08-23
feat(server): add getTunneltime to manager metrics

daniellacosse:
successful result from digital ocean droplet:

Screenshot 2024-09-11 at 3 38 22 PM

TODOs:

- [x] figure out how to actually send traffic through the endpoint locally
- [x] set up server in digital ocean and test it that way
- [x] write test

### DIFF
--- a/src/shadowbox/server/api.yml
+++ b/src/shadowbox/server/api.yml
@@ -418,6 +418,35 @@ paths:
               examples:
                 '0':
                   value: '{"bytesTransferredByUserId":{"1":1008040941,"2":5958113497,"3":752221577}}'
+  /metrics/tunnel-time-location
+    get:
+      description: Returns the tunnel time per location
+      tags:
+        - Tunnel time
+      responses:
+        '200':
+          description: The tunnel time per location
+          content: 
+            application/json:
+              schema:
+                type: array
+                items:
+                  type: object
+                  properties:
+                    location:
+                      type: string
+                    asn: 
+                      type: number
+                    as_org:
+                      type: string
+                    tunnel_time:
+                      type: object
+                      properties:
+                        seconds:
+                          type: number
+              examples:
+                '0':
+                  value: '[{"location":"US","asn":7922,"as_org":"Comcast Cable Communications, LLC","tunnel_time":{"seconds":2523432}]'
   /metrics/enabled:
     get:
       description: Returns whether metrics is being shared
--- a/src/shadowbox/server/manager_metrics.ts
+++ b/src/shadowbox/server/manager_metrics.ts
@@ -15,24 +15,26 @@
 import {PrometheusClient} from '../infrastructure/prometheus_scraper';
 import {DataUsageByUser, DataUsageTimeframe} from '../model/metrics';
 
-interface TunnelTimeDuration {
+interface Duration {
   seconds: number;
 }
 
 interface TunnelTimeRequest {
-  time_window: TunnelTimeDuration;
+  time_window: Duration;
 }
 
-interface TunnelTimeResponse {
+interface TunnelTimeResponseEntry {
   location?: string;
   asn?: number;
   as_org?: string;
-  tunnel_time: TunnelTimeDuration;
+  tunnel_time: Duration;
 }
+
+type TunnelTimeResponse = TunnelTimeResponseEntry[];
 
 export interface ManagerMetrics {
   getOutboundByteTransfer(timeframe: DataUsageTimeframe): Promise<DataUsageByUser>;
-  getTunnelTimeByLocation(request: TunnelTimeRequest): Promise<TunnelTimeResponse[]>;
+  getTunnelTimeByLocation(request: TunnelTimeRequest): Promise<TunnelTimeResponse>;
 }
 
 // Reads manager metrics from a Prometheus instance.
@@ -57,11 +59,9 @@ export class PrometheusManagerMetrics implements ManagerMetrics {
     return {bytesTransferredByUserId: usage};
   }
 
-  async getTunnelTimeByLocation({
-    time_window: {seconds},
-  }: TunnelTimeRequest): Promise<TunnelTimeResponse[]> {
+  async getTunnelTimeByLocation(request: TunnelTimeRequest): Promise<TunnelTimeResponse> {
     const {result} = await this.prometheusClient.query(
-      `sum(increase(shadowsocks_tunnel_time_seconds_per_location[${seconds}s])) by (location, asn, asorg)`
+      `sum(increase(shadowsocks_tunnel_time_seconds_per_location[${request.time_window.seconds}s])) by (location, asn, asorg)`
     );
 
     return result.map((entry) => ({
--- a/src/shadowbox/server/manager_service.ts
+++ b/src/shadowbox/server/manager_service.ts
@@ -157,7 +157,10 @@ export function bindService(
   );
 
   apiServer.get(`${apiPrefix}/metrics/transfer`, service.getDataUsage.bind(service));
-  apiServer.get(`${apiPrefix}/metrics/tunnel-time`, service.getTunnelTimeByLocation.bind(service));
+  apiServer.get(
+    `${apiPrefix}/metrics/tunnel-time-location`,
+    service.getTunnelTimeByLocation.bind(service)
+  );
   apiServer.get(`${apiPrefix}/metrics/enabled`, service.getShareMetrics.bind(service));
   apiServer.put(`${apiPrefix}/metrics/enabled`, service.setShareMetrics.bind(service));
 


src/shadowbox/model/metrics.ts
```diff
@@ -28,3 +28,15 @@ export interface DataUsageByUser {
...
 }
+
+
+export interface TunneltimeMetrics {
+  access_key?: Map<string, number>;
+  country?: Map<string, number>;
+  asn?: Map<string, number>;
+}
+
+export interface TunneltimeQuery {
```
fortuna, line 39: We need to define a TunnelTimeRequest and TunnelTimeResponse object. Let's call it that. I'd say prefer request over query because it's more generic and aligned with the practices at Google.

Request and response are service concepts, and should not be in the domain model.

We will need to define the domain model concepts that we expose in the UI, but they must be decoupled.

fortuna, line 39: The request shape looks mostly good. I would use a duration type, and not call it hours. Perhaps `time_window` or `since_time_ago` kind of concept.

For the group name. Replace it with a list of group by fields. Alternatively we can return all dimensions always, but i like the idea of specifying the group by dimensions, as it can be a big optimization for large servers.

src/shadowbox/model/metrics.ts
```diff
@@ -28,3 +28,15 @@ export interface DataUsageByUser {
 export interface DataUsageTimeframe {
   hours: number;
 }
+
+
+export interface TunneltimeMetrics {
```
fortuna, line 33: The response should contain a list of values. Each value is a duration, with dimensions. The client code can then iterate over the entries, aggregating accordingly.

Example:
```yml
[
  {key: key1, country: IR, asn: 1234, tunnel_time: {hours: 10}},
  {key: key1, country: IR, asn: 5678, tunnel_time: {hours: 2}}
]
```

fortuna (changes requested):
Thanks for the PR. This is exactly the kind of discussion I was looking for!

src/shadowbox/server/manager_metrics.ts
```diff
@@ -15,8 +15,21 @@
...
+export type TunnelTimeDimension = 'access_key' | 'country' | 'asn';
+
+interface TunnelTimeRequest {
+  hours: number;
+}
+
+interface TunnelTimeResponse {
+  location?: string;
+  asn?: number;
+  tunnel_time: number;
```
sbruens, line 27: Use a suffix to indicate the unit of duration. Loosely following https://google.aip.dev/142 I'm inclined to suggest `tunnel_time_seconds`, using seconds instead of hours as the unit.

src/shadowbox/server/mocks/mocks.ts
```diff
@@ -65,3 +65,21 @@ export class FakePrometheusClient extends PrometheusClient {
     return queryResultData;
   }
 }
+
+export class FakeTunnelTimePrometheusClient extends PrometheusClient {
```
sbruens, line 69: Is it worth renaming the existing `FakePrometheusClient` to something like `FakeDataBytesTransferredPrometheusClient`?

sbruens (approved):

src/shadowbox/server/manager_metrics.ts
```diff
@@ -15,8 +15,21 @@
...
 
+export type TunnelTimeDimension = 'access_key' | 'country' | 'asn';
+
+interface TunnelTimeRequest {
+  hours: number;
+}
+
+interface TunnelTimeResponse {
+  location?: string;
+  asn?: number;
```
fortuna, line 26: We should return as_org as well, now that we have that. We will be using that in the UI

src/shadowbox/server/manager_metrics.ts
```diff
@@ -15,8 +15,21 @@
...
+  tunnel_time_seconds: number;
```
fortuna, line 27: This is inconsistent with the way we do time in other places in the API. We use something like:

```ts
{
  tunnel_time: {
    seconds: number
  }
}
```

src/shadowbox/server/manager_service.ts
```diff
@@ -156,6 +163,10 @@ export function bindService(
   );
 
   apiServer.get(`${apiPrefix}/metrics/transfer`, service.getDataUsage.bind(service));
+  apiServer.get(
+    `${apiPrefix}/metrics/tunnel/location`,
```
fortuna, line 167: This is not proper REST language. You are retrieving tunnel time, not locations.

I'd call this `/metrics/tunnel_time`.

Also, why breakdown by location only? I suggest we return all the dimensions by default. Or specify the dimensions we want in the request.

src/shadowbox/server/manager_service.ts
```diff
@@ -24,6 +24,7 @@ import {AccessKey, AccessKeyRepository, DataLimit} from '../model/access_key';
 import * as errors from '../model/errors';
 import * as version from './version';
 
+import type {TunnelTimeDimension} from './manager_metrics';
```
fortuna, line 27: This is not being used.

src/shadowbox/server/manager_service.ts
```diff
@@ -75,10 +76,16 @@ interface RequestParams {
...
+
+interface RequestQuery {
+  [param: string]: unknown;
+}
+
 // Simplified request and response type interfaces containing only the
 // properties we actually use, to make testing easier.
 interface RequestType {
   params: RequestParams;
+  query?: RequestQuery;
```
fortuna, line 88: Is this used?

fortuna (changes requested):

src/shadowbox/server/manager_metrics.ts
sbruens: Ugh I didn't realize we already use this format elsewhere. I suppose we could continue using it for consistency's sake, but it's an odd pattern to represent time in an API.

At least `seconds` is better than `hours`, so I guess that's something? It gives us the option to add nanoseconds in future and be more similar to the [`Timestamp` proto](https://github.com/protocolbuffers/protobuf/blob/6b22fd125ce20f56e89d0f207e36383d19a103fd/src/google/protobuf/timestamp.proto#L133). Let's not use it with anything other than `seconds` though, and we can consider fixing this in some unknown future.

fortuna: Note that this is not a timestamp. This is a duration. I could imagine a Duration type that can have whatever unit we want, and a function that goes from JSON to a Duration object. (I think we already do that?)

src/shadowbox/server/manager_metrics.ts
```diff
@@ -15,8 +15,22 @@
 import {PrometheusClient} from '../infrastructure/prometheus_scraper';
 import {DataUsageByUser, DataUsageTimeframe} from '../model/metrics';
 
+export type TunnelTimeDimension = 'access_key' | 'country' | 'asn';
+
+interface TunnelTimeRequest {
+  hours: number;
```
fortuna, line 21: This should probably use the same duration object as the Response.

Also, it helps to use a name that represents what it is, like `window_length` or `time_window`?

src/shadowbox/server/manager_metrics.ts
```diff
@@ -40,4 +58,21 @@ export class PrometheusManagerMetrics implements ManagerMetrics {
     }
     return {bytesTransferredByUserId: usage};
   }
+
+  async getTunnelTimeByLocation({
```
fortuna, line 62: This is no longer by location. It's broken down by all dimensions. Please rename.

src/shadowbox/server/manager_metrics.ts
```diff
@@ -15,8 +15,26 @@
 import {PrometheusClient} from '../infrastructure/prometheus_scraper';
 import {DataUsageByUser, DataUsageTimeframe} from '../model/metrics';
 
+export type TunnelTimeDimension = 'access_key' | 'country' | 'asn';
+
+interface TunnelTimeRequest {
+  time_window: {
```
fortuna, line 21: Let's define a Duration type and reuse

src/shadowbox/server/manager_metrics.ts
```diff
@@ -40,4 +58,21 @@ export class PrometheusManagerMetrics implements ManagerMetrics {
...
+    const {result} = await this.prometheusClient.query(
+      `sum(increase(shadowsocks_tunnel_time_seconds_per_location[${seconds}s])) by (location, asn, asorg)`
+    );
+
+    return result.map((entry) => ({
+      location: entry.metric['location'],
+      asn: entry.metric['asn'] !== undefined ? parseInt(entry.metric['asn'], 10) : undefined,
+      as_org: entry.metric['asorg'],
+      tunnel_time: {
+        seconds: Math.round(parseFloat(entry.value[1])),
```
fortuna, line 74: Let's not round here.

src/shadowbox/server/manager_service.ts
```diff
@@ -599,6 +601,23 @@ export class ShadowsocksManagerService {
     }
   }
 
+  async getTunnelTimeByLocation(req: RequestType, res: ResponseType, next: restify.Next) {
```
fortuna, line 604: Rename

src/shadowbox/server/manager_metrics.spec.ts
```diff
@@ -27,4 +30,16 @@ describe('PrometheusManagerMetrics', () => {
     expect(bytesTransferredByUserId['access-key-2']).toEqual(10000);
     done();
   });
+
+  it('getTunnelTimeByLocation', async (done) => {
```
fortuna, line 34: Let's have an example that shows that we can have multiple entries for each  country, asn or as_org.

src/shadowbox/server/manager_metrics.ts
fortuna: It would be:

```ts
interface Duration {
   seconds: number;
}
```

Only one field. This makes the type reusable, and clearly marks all the variables that take a duration.

src/shadowbox/server/manager_metrics.ts
```diff
@@ -15,8 +15,26 @@
...
+  };
+}
+
+interface TunnelTimeResponse {
+  location?: string;
+  asn?: number;
+  as_org?: string;
+  tunnel_time: {
+    seconds: number;
+  };
```
fortuna, line 32:
```suggestion
  tunnel_time: Duration;
```

src/shadowbox/server/manager_metrics.ts
```diff
@@ -15,8 +15,26 @@
 import {PrometheusClient} from '../infrastructure/prometheus_scraper';
 import {DataUsageByUser, DataUsageTimeframe} from '../model/metrics';
 
+export type TunnelTimeDimension = 'access_key' | 'country' | 'asn';
+
+interface TunnelTimeRequest {
+  time_window: {
+    seconds: number;
+  };
```
fortuna, line 23:
```suggestion
  time_window: Duration;
```

src/shadowbox/server/manager_metrics.ts
```diff
@@ -15,8 +15,24 @@
...
 
+interface TunnelTimeDuration {
+  seconds: number;
+}
+
+interface TunnelTimeRequest {
+  time_window: TunnelTimeDuration;
+}
+
+interface TunnelTimeResponse {
```
fortuna, line 26: The naming here is a bit off. A request has only one response. The TunnelTime response is actually a list of entries.

Perhaps call this TunnelTimeEntry?

src/shadowbox/server/manager_metrics.ts
```diff
@@ -15,8 +15,24 @@
 import {PrometheusClient} from '../infrastructure/prometheus_scraper';
 import {DataUsageByUser, DataUsageTimeframe} from '../model/metrics';
 
+interface TunnelTimeDuration {
```
fortuna, line 18: I think we can call this simply `Duration`.

src/shadowbox/server/manager_metrics.ts
```diff
@@ -40,4 +56,21 @@ export class PrometheusManagerMetrics implements ManagerMetrics {
     }
     return {bytesTransferredByUserId: usage};
   }
+
+  async getTunnelTimeByLocation({
+    time_window: {seconds},
```
fortuna, line 61: To ensure it's the proper type here
```suggestion
    time_window: Duration,
```

fortuna (commented):
Looks good. Just some minor tweaks. And we need to update the YAML file.

src/shadowbox/server/api.yml
```diff
@@ -418,6 +418,35 @@ paths:
               examples:
                 '0':
                   value: '{"bytesTransferredByUserId":{"1":1008040941,"2":5958113497,"3":752221577}}'
+  /metrics/tunnel-time-location
```
fortuna, line 421: Can this be `/metrics/tunnel-time`? I thought you mention there was another one per access key, but I don't see it. How are we going to report by key?

Did Sander have an opinion on this?

fortuna: We will know better about the API once we start using it for real. So perhaps we should pause this and focus on the manager side to better understand the usage.

For instance, we need tunnel time per access key too. Will the manager be forced to issue 3 queries?
Perhaps it will be best to issue one query, maybe under `/metrics/usage`, that returns all the usage metrics we need. That will simplify the Manager. It will also let us deprecate the bad `transfer` endpoint. In the future we can just add more to the usage endpoint, instead of keeping adding new endpoints per metric.

PR #1626, opened 2024-12-17
fix(server): fix and validate open API spec

daniellacosse:
Screenshot 2024-12-17 at 11 19 34 AM


sbruens (approved):
Nice, thanks for fixing the other issues as well!

PR #1637, opened 2025-02-12
ACCESS_KEY_PORT are not correctly configured

terencekam:


sbruens:
Thanks for the pull request, @terencekam.

This script is no longer maintainted. The up-to-date version lives in https://github.com/Jigsaw-Code/outline-apps/blob/master/server_manager/install_scripts/install_server.sh. Any changes we make should be applied to that new location.

I'm trying to understand the issue you're trying to fix. Can you elaborate on what you're experiencing?

PR #1690, opened 2026-01-05
fix(server): switch to maintained watchtower fork

ohnorobo:
Fix for https://github.com/OutlineFoundation/outline-server/issues/1684

[`containrrr/watchtower`](https://github.com/containrrr/watchtower) is no longer maintained, and has a breaking issue at https://github.com/containrrr/watchtower/issues/2122

https://github.com/nicholas-fedor/watchtower is the new maintained fork the community is moving to.

### DIFF
--- a/src/server_manager/install_scripts/install_server.sh
+++ b/src/server_manager/install_scripts/install_server.sh
@@ -376,7 +376,7 @@ function start_watchtower() {
       -v /var/run/docker.sock:/var/run/docker.sock)
   # By itself, local messes up the return code.
   local STDERR_OUTPUT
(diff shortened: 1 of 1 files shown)


fortuna (changes requested):
If we are going to change the image, we should at a minimum [introduce a redirector](https://httptoolkit.com/blog/docker-image-registry-facade/) to allow for future migrations without disruption.

In any case, the right place to fix this is in the outline-apps repo, where the Manager lives. This location is deprecated. It was kept for backwards-compatibility. We might be able to delete it at this point.

Issue #1039, opened 2022-01-22
[Proposal] Supporting arch other than x86_64 or amd64 (with how)


seia-soto:
Hello? I recently created a simple script that builds multi-platform supported docker image of shadowbox (or outline-server).
During building the script, I got some clean ways to build multi-platform shadowbox image even there is directory called `third_party` which is platform-sensitive.

## Using Docker BuildKit's `$TARGETPLATFORM`

Thanks to Docker BuildKit's `TARGETPLATFORM` argument in `Dockerfile`, we can use it in `COPY` command inside of `Dockerfile`.

Like following snippet:

```dockerfile
ARG TARGETPLATFORM
RUN mkdir -p third_party
COPY third_party/outline-ss-server/${TARGETPLATFORM} third_party/outline-ss-server
COPY third_party/prometheus/${TARGETPLATFORM} third_party/prometheus
```

> Note that ARG has scoping issue, so we need to put it after `FROM` keyword.

Following to above, we can expect paths like `third_party/${TARGETPLATFORM}`:

- third_party/linux/arm64
- third_party/linux/amd64
- third_party/linux/arm/v7
- third_party/linux/arm/v6
- ...

Finally, we can change current `third_party` directory structure into **(only considering docker implementation)**:

```
- `${TARGETPLATFORM}`
  - outline-ss-server (binary)
  - prometheus (binary)

// respect current directory structures
- outline-ss-server
  - LICENSE
  - METADATA
- prometheus
  - LICENSE
  - METADATA
- shellcheck (keep original as it is not required at build process)
```

> I agree that this implementation is not completed and should not be used as only works for docker.

## Using `$(uname -m)`

Or if we target both docker way and non-docker way building, I think it's reasonable to change directory structure into:

```bash
- $([[ "$(uname -s)" == "Darwin" ]] && echo "macos" || echo "linux")
  - $(uname -m) # For ARM, it should be arm64 if Apple based, otherwise aarch64
    - outline-ss-server
    - prometheus

: '
- linux
  - x86_64
  - aarch64
  - aarch32? (I could not check it)
- macos
  - x86_64
  - arm64
'
```

- [StackOverflow: Answer of Differences between arm64 and aarch64](https://stackoverflow.com/a/31852003)

Also, modify the build script.

> We still need additional changes for docker builds.

```bash
# Install third_party dependencies
readonly OS="$([[ "$(uname)" == "Darwin" ]] && echo "macos" || echo "linux")"
[...]

PR #1700, opened 2026-03-25
chore(server): port shadowbox Docker image to arm64 

oceanapplications:
Added support to run the server on Apple Silicon MacOS and arm64 Linux.

### DIFF
--- a/.github/workflows/build_and_test_debug.yml
+++ b/.github/workflows/build_and_test_debug.yml
@@ -96,7 +96,7 @@ jobs:
           cache: npm
 
       - name: Install NPM Dependencies
-        run: npm ci
+        run: PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true npm ci
 
       - name: Shadowbox Debug Build (arm64)
         run: TARGET_ARCH=aarch64 ./task shadowbox:build
--- a/.github/workflows/build_and_test_debug.yml
+++ b/.github/workflows/build_and_test_debug.yml
@@ -46,7 +46,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v5
         with:
-          go-version: '1.21'
+          go-version-file: go.mod
           cache: false
 
       - name: Install NPM Dependencies
@@ -106,7 +106,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v5
         with:
-          go-version: '1.21'
+          go-version-file: go.mod
           cache: false
 
       - name: Install NPM Dependencies
--- a/.github/workflows/build_and_test_debug.yml
+++ b/.github/workflows/build_and_test_debug.yml
@@ -43,8 +43,16 @@ jobs:
           node-version: 18
           cache: npm
 
+      - name: Install Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.21'
+          cache: false
+
       - name: Install NPM Dependencies
         run: npm ci
+        env:
+          PUPPETEER_SKIP_CHROMIUM_DOWNLOAD: 'true'
 
       - name: Lint
         run: ./task lint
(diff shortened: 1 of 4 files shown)


ohnorobo:
Hi @oceanapplications , thanks for your interest in adding support for other platforms in Outline. arm64 support is definitely something we'd be interested in adding.

Currently the CI is not passing for the changes you added. Could you take a look at that? You can also take a look at https://github.com/OutlineFoundation/outline-server/issues/1039 to see some prior work on supporting other platforms.

oceanapplications:
> Hi @oceanapplications , thanks for your interest in adding support for other platforms in Outline. arm64 support is definitely something we'd be interested in adding.
> 
> Currently the CI is not passing for the changes you added. Could you take a look at that? You can also take a look at #1039 to see some prior work on supporting other platforms.

I’ll it passing. It’s an issue with official chromium being unavailable for arm64 but I think there’s plenty of workarounds.

I’ve been running my fork on arm64 for a couple weeks with no issues so it’s something I think we should be able to support.

ohnorobo:
FYI, we don't auto-run github actions on external pull requests due to some security issues with github actions in the past. It would probably be faster for you to also make an internal PR within your fork so you can testrun the CI without being blocked on me approving the CI run. (This was the source of the delay in getting to this review earlier, sorry about that.)

oceanapplications:
Passing on my branch now. Pushed over outline server running on arm64 Raspberry Pi.

.github/workflows/build_and_test_debug.yml
```diff
@@ -43,8 +43,16 @@ jobs:
           node-version: 18
           cache: npm
 
+      - name: Install Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.21'
```
fortuna: This is a maintenance burden. Please refer to the go.mod file like we do here: https://github.com/OutlineFoundation/outline-sdk/blob/0cf8e09b67e8e7b5930554f1298920ac7e079f34/.github/workflows/test.yml#L29

Same below

.github/workflows/build_and_test_debug.yml
```diff
@@ -81,6 +89,40 @@ jobs:
...
+      - name: Install Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          cache: npm
+
+      - name: Install Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.21'
```
fortuna: Use go.mod file

src/server_manager/install_scripts/install_server.sh
fortuna, line 1: I believe we no longer use this script. You will have to update it in outline-apps as a separate PR

src/shadowbox/integration_test/test.sh
```diff
@@ -100,10 +100,13 @@ function setup() {
...
 
   # Shadowsocks service.
+  # Start on NET_OPEN first so that -p host port binding works on macOS Docker Desktop
+  # (Docker Desktop does not publish ports when the initial network is --internal).
+  # Then connect to NET_BLOCKED so the security isolation tests still pass.
   declare -ar shadowbox_flags=(
     -d
     --rm
-    --network="${NET_BLOCKED}"
+    --network="${NET_OPEN}"
```
fortuna: This is modifying the test. It's not longer in the blocked network, so we won't be testing that it bypasses the block. This needs to run on the blocked network.

fortuna (changes requested):

src/server_manager/install_scripts/install_server.sh
oceanapplications: Should I leave the changes in here for consistency? I can also add a PR to outline-apps.

src/shadowbox/integration_test/test.sh
oceanapplications: I reverted the test so it's simple and not confusing. That other stuff was to get tests working locally on my Mac which probably isn't necessary.

src/server_manager/install_scripts/install_server.sh
ohnorobo: I would update both for now since it's also used as part of the CI here. We may separately decide to get rid of it but it probably requires some CI detangling that is not your problem to solve.

ohnorobo (approved):

fortuna (approved):

PR #1704, opened 2026-05-11
chore: upgrade lib versions for quay security scanner

### LINKED ISSUE #1703: Are you tracking security issues pertaining to dependencies you're using?
Hi,

I noticed on [quay.io](https://quay.io/repository/outline/shadowbox) that the latest image seems to contain several security issues, ranging up to "critical," obviously due to dependencies you're using. I can't tell whether these are applicable to Outline at all.

I just wanted to understand whether anybody is looking into these reports and taking action, if and when required.

Many thanks.

ohnorobo:
Upgrading versions to cover the https://quay.io/repository/outline/shadowbox/manifest/sha256:545c6f7c7261bb30ae1dffe24a6fca5f8512f5d17c72cfb9e410e7e655444e62?tab=vulnerabilities security scanner issues.

None of these are actually critical for us (we don't use `ServerConfig.PublicKeyCallback`, etc) but it's still a good idea to upgrade.

Doing this now since we'll want to do a release to pick up https://github.com/OutlineFoundation/outline-server/pull/1700 anyway.

Fixes https://github.com/OutlineFoundation/outline-server/issues/1703


go.mod
```diff
@@ -1,6 +1,8 @@
 module localhost
 
-go 1.21
+go 1.24.0
```
fortuna: Perhaps upgrade to the latest here. It's the safest option.

ohnorobo: updated to latest stable, 1.26.3

fortuna (approved):

PR #1705, opened 2026-05-11
chore: lib upgrades for trivy issues

ohnorobo:
Further upgrades for lib version security issues surfaced by [Trivy](https://trivy.dev/)

- upgrades Prometheus 2.53.4 → 3.5.3 (LTS)
- removes unneeded npm from the docker image.

### DIFF
--- a/src/shadowbox/docker/Dockerfile
+++ b/src/shadowbox/docker/Dockerfile
@@ -28,7 +28,10 @@ STOPSIGNAL SIGKILL
 # Upgrade installed Alpine packages to pick up security fixes (musl-utils, busybox, ssl_client, etc.).
 # We use curl to detect the server's public IP. We need to use the --date option in `date` to
 # safely grab the ip-to-country database.
-RUN apk upgrade --no-cache && apk add --no-cache --upgrade coreutils curl
+# Remove npm/npx CLI — the runtime only invokes `node`, and npm ships old transitive deps that
+# trip image vulnerability scanners.
+RUN apk upgrade --no-cache && apk add --no-cache --upgrade coreutils curl \
+  && rm -rf /usr/local/lib/node_modules/npm /usr/local/bin/npm /usr/local/bin/npx
 
 COPY . /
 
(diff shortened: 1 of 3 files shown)


third_party/Taskfile.yml
```diff
@@ -11,16 +11,16 @@ tasks:
   prometheus:download-*-*:
     desc: Download and extract prometheus binary
     vars:
-      VERSION: '2.53.4'
+      VERSION: '3.5.3'
```
fortuna, line 14: This is a major version change. Will it be able to read the v2 data? If so, we need a migration step.

src/shadowbox/docker/Dockerfile
```diff
@@ -28,7 +28,10 @@ STOPSIGNAL SIGKILL
 # Upgrade installed Alpine packages to pick up security fixes (musl-utils, busybox, ssl_client, etc.).
 # We use curl to detect the server's public IP. We need to use the --date option in `date` to
 # safely grab the ip-to-country database.
-RUN apk upgrade --no-cache && apk add --no-cache --upgrade coreutils curl
+# Remove npm/npx CLI — the runtime only invokes `node`, and npm ships old transitive deps that
+# trip image vulnerability scanners.
```
fortuna, line 32: BTW, we don't have to over-index in vulnerability scanners if they are not exploitable. There's a ton of false-positives. I would imagine that a new node image would address that. Why don't you upgrade the node image instead?

third_party/Taskfile.yml
ohnorobo: Ah yeah, it does require at least thinking through the implications of https://prometheus.io/docs/prometheus/latest/migration/

ohnorobo:
Yeah, I split this out because it came up while I was working on https://github.com/OutlineFoundation/outline-server/pull/1704 but I agree that there's no pressing need to do these upgrades. Particularly if they require deeper migration planning we can drop this one for now.

